PR #38, opened 2021-09-26
Fix building of e.g., vendor/tree-sitter-haskell-src/scanner.cc

bbannier:
At least this parser uses C++11 (lambda expressions) and C++14 features
(lambda `auto` parameter types). Since the default C++ standard of
compilers can vary, this patch explicitly sets C++14 mode when building
any vendored C++ files.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,6 +27,7 @@ impl TreeSitterParser {
             cpp_build
                 .include(&dir)
                 .cpp(true)
+                .flag("--std=c++14")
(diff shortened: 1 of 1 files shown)


Wilfred:
Thanks! What was the platform where you noticed this issue?

bbannier:
I ran into this on macos-10.15.7 with I believe the system Clang (`Apple clang version 12.0.0 (clang-1200.0.32.29)`).

PR #41, opened 2021-10-01
Support `cjs` and `mjs` extensions for JavaScript

cherryblossom000:
This PR adds `cjs` and `mjs` to the extensions matched for JavaScript. Node.js uses the [`cjs` and `mjs` extensions][1] to indicate different module systems.

[1]: https://nodejs.org/api/packages.html#packages_determining_module_system


Wilfred:
Thanks :)

Issue #100, opened 2022-01-18
Perform slider alignment after diffing


Wilfred:
This would simplify the tree diff considerably. At each level of the tree, look for novel sequences where the the first novel node is equal to the following unchanged node.

Wilfred:
Done in 3548c32a4.

PR #101, opened 2022-01-18
Add Scala support

hugo-vrijswijk:
I've mostly followed [this guide](https://difftastic.wilfred.me.uk/upstream_parsers.html) to add [tree-sitter-scala](https://github.com/tree-sitter/tree-sitter-scala). I don't have a full grasp of what `atom_nodes` includes, so diffing for that might not be perfect, but I think it's better than nothing. There's no `highlight_query` as tree-sitter-scala doesn't have a `highlights.scm`.

Any changes from me are in https://github.com/Wilfred/difftastic/pull/101/commits/ebd711bae20f0a041dd5206fe1cb8e32f1d878a5

I've also added two test files which seem to give a nice diff:

image


src/tree_sitter_parser.rs
```diff
@@ -379,6 +380,16 @@ pub fn from_language(language: guess::Language) -> TreeSitterConfig {
                 .unwrap(),
             }
         }
+        Scala => {
+            let language = unsafe { tree_sitter_scala() };
+            TreeSitterConfig {
+                name: "Rust",
```
Wilfred, line 386:
```suggestion
                name: "Scala",
```

You probably want to change this :)

Wilfred:
Thanks for doing this! Other that one copy-paste error, this looks good. Would you mind linking to the grammar you've used in `manual/src/upstream_parsers.md`?

src/tree_sitter_parser.rs
hugo-vrijswijk: Oops. Thanks for catching it!

Wilfred:
Thanks!

FWIW performance on master is significantly worse than 0.15 currently, so difftastic will probably struggle with diffing larger Scala files until #100 is done.

PR #205, opened 2022-03-30
Add gleam

jbr:
Thanks!


jbr:
I don't really understand the reordering in the output regression test — I ran `./sample_files/compare_all.sh; cp sample_files/compare.result sample_files/compare.expected` and there was an order change. I'm on macos 12.3 with bash 3.2.57, but I don't see why that would change anything about the order of file iteration inside the script

Wilfred:
See also #211, looks like multiple people are interested in Gleam support :)

> It looks like the sort order of - and _ differ between my machine and github actions?

`compare_all.sh` explicitly sets `LANG` so that the order of files when globbing should be consistent. Looks like that isn't sufficient for macOS -- it's probably easier to just add `| sort` to make this robust.

> I wasn't sure what to put for atom_nodes and delimiter_tokens so I left them blank

I've added a new section in the manual to make this clearer: https://difftastic.wilfred.me.uk/adding_a_parser.html#configure-parsing

jbr:
@Wilfred Rebased, added atom_nodes and delimiter_tokens

PR #206, opened 2022-03-30
Fix "she-bang" typo

mjul:
This makes the test correct in case the full rule of only recognizing an Emacs mode line on the second line if the first line is a "she-bang" line is ever fully implemented.


Wilfred:
Thanks, well spotted.

PR #210, opened 2022-03-30
Fixes suggested by `cargo clippy`

Dr-Emann:
Fixes as suggested by clippy, such as reducing nesting in if-then chains, replacing `!result.is_ok()` with `result.is_err()`, using the `matches!` macro, and matching on the result of `x.cmp(y)` rather than an if-then chain of `if x == y {} else if x < y {} else {}`

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -11,7 +11,7 @@ pub const DEFAULT_BYTE_LIMIT: usize = 1_000_000;
 
 const USAGE: &str = concat!(env!("CARGO_BIN_NAME"), " [OPTIONS] OLD-PATH NEW-PATH");
 
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
 pub enum ColorOutput {
     Always,
     Auto,
@@ -314,7 +314,7 @@ pub fn parse_args() -> Mode {
 /// Choose the display width: try to autodetect, or fall back to a
 /// sensible default.
 fn detect_display_width() -> usize {
-    term_size::dimensions().map(|(w, _)| w).unwrap_or(80)
+    term_size::dimensions().map_or(80, |(w, _)| w)
 }
 
 pub fn should_use_color(color_output: ColorOutput) -> bool {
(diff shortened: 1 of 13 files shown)


src/options.rs
```diff
@@ -223,16 +223,14 @@ pub fn parse_args() -> Mode {
...
             BackgroundColor::Dark
         }
-    } else {
-        if let Ok(background) = env::var("DFT_BACKGROUND") {
-            if background == "light" {
-                BackgroundColor::Light
-            } else {
-                BackgroundColor::Dark
-            }
+    } else if let Ok(background) = env::var("DFT_BACKGROUND") {
```
jbr, line 226: It looks like this logic could be simplified even further:

```suggestion
    } else if let Ok("light") = env::var("DFT_BACKGROUND").as_deref() {
        BackgroundColor::Light
    } else {
        BackgroundColor::Dark
    }
```

Dr-Emann: I worry a little that it makes things a little less clear: right now it reads:

"If `matches` specifies background, use that. If not, and the environment variable is set, use that. Otherwise, use a default", but with that change, it reads as

"If `matches` specifies background, use that. If not, and the environment variable is set to "light", make it light, otherwise, use the default".

It's subtly different, and right now, they actually mean the same thing, but might be a surprise if another fallback is added before the default. I actually almost made the mistake of trying to apply the same change to the original `if`, but that would have actually changed behavior (if matches specified it, but dark, it could be overridden by the environment variable)

Still happy to make the change if you think it's better.

Wilfred:
Thanks for the PR! I'm afraid the difftastic source code is moving pretty quickly, and some of the issues apply to code that has changed since.

If you'd like to update this PR to fix the remaining clippy issues, we can merge this if you're still interested :)

Wilfred:
Merged in 1fb8ba7ee4e7ab50f0c7a99f3ed2675c942cc171, thank you :)

PR #214, opened 2022-03-30
Make --color affect missing line number formatting

### LINKED ISSUE #213: Not all terminal escape codes are disabled by `--color never `
So I'm playing with the tool, trying to use it with other utilities such as `tig` and `less`. Not all of these support color, and in these cases I see ANSI escape codes. When I try to disable these with `--color never` most are gone, but some are left, which I see as a problem.

Color enabled:
`echo "foo\nbar" > a; echo "baz" > b; difft --color always a b | less`
```
ESC[1mESC[93mbESC[39mESC[0m --- 1/2 --- Text
ESC[91m1 ESC[39mESC[92m1 ESC[39mESC[91;1mfooESC[0m
ESC[91m2 ESC[39mESC[2m. ESC[0mESC[91;1mbarESC[0m

ESC[1mESC[93mbESC[39mESC[0m --- 2/2 --- Text
ESC[91m1 ESC[39mESC[92m1 ESC[39mESC[92;1mbazESC[0m
ESC[2m. ESC[0mESC[2m. ESC[0m
```
Color disabled:
`echo "foo\nbar" > a; echo "baz" > b; difft --color never a b | less`
```
b --- 1/2 --- Text
1 1 foo
2 ESC[2m. ESC[0mbar

b --- 2/2 --- Text
1 1 baz
ESC[2m. ESC[0mESC[2m. ESC[0m
```

So ANSI escapes around the dot which is used when line number is not available are not disabled with `--color never`, which they should.

AMDmi3:
Fixes #213

### DIFF
--- a/src/side_by_side.rs
+++ b/src/side_by_side.rs
@@ -37,6 +37,7 @@ fn format_missing_line_num(
     prev_num: LineNumber,
     source_dims: &SourceDimensions,
     is_lhs: bool,
+    use_color: bool,
 ) -> String {
     let column_width = if is_lhs {
         source_dims.lhs_line_nums_width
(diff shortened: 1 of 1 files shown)


Wilfred:
Thanks for the great bug report and PR :)

PR #218, opened 2022-03-31
Add Zig support

Andoryuuta:
Adds Zig support using the [zig tree sitter implementation](https://github.com/maxxnino/tree-sitter-zig) from @maxxnino

~~PR marked as draft pending license clarification (https://github.com/maxxnino/tree-sitter-zig/issues/13).~~ Author added MIT license to project.

Before:
image

After:
image


Wilfred:
Thanks @Andoryuuta :)

By the way, it looks like you aren't getting much syntax highlighting because Zig is using `@keyword.foo` rather than just `@keyword` in its `queries.scm`. This is probably better to fix upstream than in difftastic.

PR #222, opened 2022-03-31
simplify creation of atom nodes & delimiter tokens

jbr:
While I was in tree_sitter_parser for #205 I noticed that a bunch of parsers were collecting a vec to a vec (`vec![("(", ")"].into_iter().collect::<Vec<(&'static str, &'static str)>>()`, so I removed the into_iter().collect(). As long as I was doing that, I also changed the `vec!["hashset"].into_iter().collect()` into `["hashset"].into()` for brevity. Thanks!


Wilfred:
Thank you :)

PR #225, opened 2022-04-01
Add Swift support

cherryblossom000:
Adds support for Swift. [tree-sitter-swift does not include `parser.c` in its repository](https://github.com/alex-pinkus/tree-sitter-swift#where-is-your-parserc), so I generated the `parser.c` using `tree-sitter-cli` and checked that in.


Wilfred:
This is looking pretty good! Would you mind making two changes?

* Fix the build. Looks like `cargo publish` is currently failing.
* Use the existing tree-sitter version. I want to benchmark tree-sitter before upgrading, as it's sometimes a performance bottleneck for difftastic.

cherryblossom000:
The build failure was due to cargo not including the `vendor/tree-sitter-swift` directory, so I symlinked `vendor/tree-sitter-swift/grammar.js` to `vendor/tree-sitter-swift-grammar.js`. This kind of feels hacky but I'm not sure how else to fix it. I also had to change the build script to output `parser.c` to `$OUT_DIR`.

I forgot to mention before that the generating of `parser.c` also takes quite some time and the build now takes ~5 minutes.

cherryblossom000:
Sorry for all the force pushes. I've just rebased the branch because of a merge conflict.

Wilfred:
Gosh, 5 minutes is a pretty long build time, especially when rust isn't very good at incremental builds in build.rs. Would you mind checking in the built file instead?

Wilfred:
Merged, thank you!

It's a little slow to start on my machine: I'm seeing 1.6 seconds for tree-sitter to load the `highlights/swift.scm`. I'm not sure why, I think this is an issue in tree-sitter itself.

Everything else looks great though :)

alex-pinkus:
I suspect the slowness is because @cherryblossom000 generated the grammar without the CLI option `--abi 14`, preventing us from taking advantage of some precomputed core IDs. That optimization reduces query load time to the ~300ms range. It should be pretty easy to rerun `tree-sitter` with that flag and check in the resulting `parser.c`.

cherryblossom000:
@alex-pinkus I used `tree-sitter generate --no-bindings --abi latest`, so that’s not the issue. The [`parser.c`](https://raw.githubusercontent.com/Wilfred/difftastic/4db120449daec84b13d0d303675b975f39274677/vendor/tree-sitter-swift/src/parser.c) also has `#define LANGUAGE_VERSION 14`, confirming it’s using v14.

alex-pinkus:
You are absolutely right, I was wrong! I still assume folks don’t know about the option because it’s new, but I should really have checked the source instead of assuming :|. My apologies!

PR #226, opened 2022-04-01
Add Elvish support

cherryblossom000:
Adds support for [Elvish](https://elv.sh/).


Xuanwo:
`Output Regression Test (pull_request)` failed, maybe we need to update the expected result file?

Wilfred:
Thanks :)

PR #263, opened 2022-04-22
feat: Add perl support

### LINKED ISSUE #250: Feature request: Perl support
First, thank you for this amazing tool.

It would be great to have Perl support even though it's not a very popular language anymore.

Xuanwo:
This PR will add Perl support via [tree-sitter-perl](https://github.com/ganezdragon/tree-sitter-perl), thanks @ganezdragon!

Fix #250


src/tree_sitter_parser.rs
```diff
@@ -473,6 +474,16 @@ pub fn from_language(language: guess::Language) -> TreeSitterConfig {
                 .unwrap(),
             }
         }
+        Perl => {
+            let language = unsafe { tree_sitter_perl() };
+            TreeSitterConfig {
+                name: "Perl",
+                language,
+                atom_nodes: vec!["string"].into_iter().collect(),
```
Wilfred, line 482: This isn't quite right. If you look at the diff on the sample files, it's not highlighting `'yes'` correctly.

Screenshot from 2022-04-22 01-05-00

It looks to me like you want `string_single_quoted` and `string_double_quoted` as atoms.

Xuanwo: Oh, I just copied them from `json`, let me check it out.

Xuanwo: Fixed.

Wilfred:
Thanks :)

PR #264, opened 2022-04-22
feat: Add HCL support

### LINKED ISSUE #238: Support for HCL format
It would be nice if Terraform and other file formats based on HCL to be supported as well.

Xuanwo:
This PR will add support for HCL via [tree-sitter-hcl](https://github.com/MichaHoffmann/tree-sitter-hcl). Thank @MichaHoffmann, for the hard work to make it possible! :rocket: 

Fix #238


Wilfred (changes requested):
Basically looks good, but you still need to set atom nodes.

Wilfred:
There's some information on atom nodes at https://difftastic.wilfred.me.uk/adding_a_parser.html#configure-parsing, but I think you want `string_lit` and `heredoc_template`.

Xuanwo:
> There's some information on atom nodes at https://difftastic.wilfred.me.uk/adding_a_parser.html#configure-parsing, but I think you want `string_lit` and `heredoc_template`.

`heredoc_template` seems should not be treated as atoms maybe?

## With `heredoc_template`

image

## Without `heredoc_template`

image

Wilfred:
@Xuanwo Unfortunately difftastic doesn't support interpolated string literals today. It assumes strings are atoms.

I'm already doing the same thing in other languages, e.g. JavaScript treats `template_string` as an atom.

It would definitely be nice to fix all these cases at some point, but that's a separate task :)

Xuanwo:
> I'm already doing the same thing in other languages, e.g. JavaScript treats `template_string` as an atom.

Thanks for explaining!

PR #281, opened 2022-05-03
feat: Improve binary content guess

### LINKED ISSUE #278: Treat PDFs as binary files

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

Fix #278 

Introduce `tree_magic_mini` to read magic bytes so that we can guess file media type more accurately and extensible.

In this PR, I add checks for

- pdf
- all image types
- all audio types
- all video types
- all font types


Wilfred:
Thanks, this is a really nice solution.

PR #282, opened 2022-05-03
ci: Adopt rust-toolchain instead

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

`1.56` is specified in tests, how about using `stable` instead?

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,4 @@
 [toolchain]
 channel = "1.56"
 components = ["rustfmt"]
+profile = "minimal"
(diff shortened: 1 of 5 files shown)


Wilfred:
Looking pretty good!

I don't want CI on stable: it's easy to accidentally use newer Rust features. By pinning the Rust version, the manual can say exactly which Rust versions are guaranteed to work.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.56"
+components = ["rustfmt"]
```
Wilfred, line 3: We should probably use the minimal profile, for consistency with the previous configuration. It's also recommended for CI here: https://rust-lang.github.io/rustup/concepts/profiles.html

```suggestion
components = ["rustfmt"]
profile = "minimal"
```

Wilfred:
Thanks :)

Issue #275, opened 2022-04-29
Support 24-bit colour


Wilfred:
This would give us nicer background highlighting and multiple shades (e.g. making certain parts of a change more prominent, similar to git-split-diffs).

ramonGonzEdu:
A few points against and for this:

Pros:
Possibly custom language-based theming
Let's you change the colors only for this app.

Cons:
Reduces the ability to have a shared theme across all cli apps by using term theme.
Makes it more likely for the fallback colors to not work well.

Notes: Im not sure if you're using bright/bold colors, but if you aren't, they could be used to highlight granular changes.

:)

FranklinYu:
I’m using Base16 themes, which makes extensive use of ANSI colors. With Base16 I can keep the command unchanged, and switch theme purely via my terminal (the only right way when SSH is involved). If 24-bit color is going to be supported, please kindly provide an ANSI fallback.

Issue #265, opened 2022-04-22
Hard to distinguish syntax highlighting from diff


Hocuri:
I often get diffs like this:

![Screenshot_2022-04-22_21-07-38](https://user-images.githubusercontent.com/18012815/164778595-45c3097d-3c1a-4173-8bc1-032dc01ed8c9.png)

where it takes me quite some time to see what actually changed because the syntax highlighting is just as colorful and eye-catching as the actual diff.

Two possible solutions:
- Add an option to turn off the syntax highlighting (I actually never wished I had syntax highlighting with the normal `git diff`, while I often wished I had a diff understanding the syntax)
- Highlight the diff more, e.g. by using red/green background for the diff (that's what GitHub does)

Thank you for this great tool!

Wilfred:
I've opened #275 to track fancier colours, but I agree disabling syntax highlighting is a reasonable feature too.

AndrewKvalheim:
I find that a nice balance is to keep syntax highlighting on but apply dim (SGR 2) to the unchanged text. Then you can even use syntax highlighting on the changes and without dim they’ll visually stand out.

Issue #167, opened 2022-03-18
Provide option to render highlights in a different style 


osa1:
difftastic currently uses bold text to highlight diffs.

However, depending on the font and terminal emulator configuration, it's possible for normal and bold text to look similar. In those cases it becomes difficult to see the important parts in a difftastic output.

It would be helpful if difftastic provided an option to use e.g. italic, or underlined etc. text for highlights so that when bold and normal text look similar on a terminal we could choose to render highlighted text with e.g. underlines.

Wilfred:
difftastic uses pretty much all the colours available on a typical terminal, plus bold, italic, dimmed, and bright colours.

I don't really want to build a theming system, especially before 1.0. It's substantial extra complexity when I'm still trying to get the core logic working well.

Long term, I plan to have HTML output and I'd like to explore 24-bit colour too.

In the meantime, do you have any particular cases in mind that don't look good?

raphael-proust:
It'd be useful to be able to deactivate syntax highlighting. I've never found a convincing simultaneous use of diff-highlighting and syntax-highlighting: there are always corner cases when the one makes the other unreadable or vice-versa.

So I think it might be useful to have the `--color` option, but also `--color-diff` and `--color-syntax` to suppress some uses of colours but not all.

- `--color` is the same as before, also `--color <VAL>` is the same as `--color-diff <VAL> --color-syntax <VAL>`,
- `--color-diff <VAL>` sets the diff coloring option only (but leaves the syntax coloring), and
- `--color-syntax <VAL>` sets the syntax coloring option only (but leaves the diff coloring).

osa1:
> In the meantime, do you have any particular cases in mind that don't look good?

Sorry I missed this. I'm using a semi-bold font, and bold version of the font does not look easy to distinguish from the normal (i.e. semi-bold) version. This makes it difficult to see changes in difftastic outputs as I need to distinguish semi-bold from bold.

durka:
I have to agree that the syntax highlighting is distracting from the actual diff highlighting. Maybe an option to use dimmer colors for the syntax, or use bold/italic for the differences instead of certain syntax elements?

durka:
Maybe this has changed since the original comment, or it depends on language?

> difftastic currently uses bold text to highlight diffs.

I see bold used for syntax, like `include`.

egrieco:
[...]

PR #286, opened 2022-05-15
Use background colors to improve diff readability

egrieco:
This pull request changes the styling of difftastic output to be closer to [delta](https://github.com/dandavison/delta) and [git-split-diffs](https://github.com/banga/git-split-diffs). Using background colors to show differences is a lot easier to understand when there is syntax highlighting involved.

I have probably missed a few things that need to be addressed before a merge. Please let me know if you see omissions or issues.

Part of this pull request rendered in the new style.

This addresses issues #167, #265, and #275.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "cansi"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "185b9281d84fc234662dc95ef999e5ac4e9b7591a6e5aa54edabda3179b80b4f"
+
+[[package]]
 name = "cc"
 version = "1.0.73"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -166,6 +172,7 @@ name = "difftastic"
 version = "0.29.0"
 dependencies = [
  "atty",
+ "cansi",
  "cc",
  "clap",
  "const_format",
@@ -188,6 +195,7 @@ dependencies = [
  "typed-arena",
  "walkdir",
  "wu-diff",
+ "yansi",
 ]
 
 [[package]]
@@ -593,3 +601,9 @@ name = "wu-diff"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8e3e6735fcde06432870db8dc9d7e3ab1b93727c14eaef329969426299f28893"
+
+[[package]]
+name = "yansi"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09041cd90cf85f7f8b2df60c646f853b7f535ce68f85244eb6731cf89fa498ec"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,8 @@ owo-colors = "3.3.0"
 rpds = "0.10.0"
 wu-diff = "0.1.2"
 rayon = "1.5.2"
+yansi = "0.5.1"
+cansi = "2.1.1"
 
 [dev-dependencies]
 pretty_assertions = "1.0.0"
--- a/src/side_by_side.rs
+++ b/src/side_by_side.rs
@@ -460,9 +460,8 @@ pub fn print(
                                 // tried several other ANSI stripping libs, this one actually works
                                     - categorise_text(&line_to_print)
                                         .iter()
-                                        .map(|s| s.text)
-                                        .collect::<String>()
-                                        .len(),
+                                        .map(|s| s.end.abs_diff(s.start) as usize)
+                                        .sum::<usize>(),
                             )
                         } else {
                             (Color::Default, 0)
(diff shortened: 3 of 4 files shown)


egrieco:
Just realized, I'm not sure I handle dark mode correctly for the general line backgrounds. It should work reasonably well for the novel item highlights though.

The colors could be chosen more carefully/intentionally in general.

egrieco:
Let's see if that makes the CI pipeline happy.

egrieco:
Found an omission. Newly added files do not have the lighter background color applied properly.

egrieco:
I've actually started using this more frequently than `delta`. With the better highlighting difftastic is excellent.

I'll try to get the conflicts and omissions resolved soon, but there is a separate piece of software that I must finish first.

Wilfred:
This is really really exciting! I agree this is the right direction for the display, bringing difftastic much closer to how changes are displayed in e.g. GitHub.

Display logic has been surprisingly hard to write for difftastic. It's often been just as challenging as the core diffing. Do you have any questions / concerns / blockers?

egrieco:
Just merged in your recent changes @Wilfred. Hopefully this makes the CI pipeline happy.

I don't have any questions or concerns at the moment, though I have been collecting a few test cases where the diffing seems to be a bit less than accurate.

Also, I have not addressed my omissions above and I'm sure there a few more that I have missed. Please let me know if you notice any obvious omissions in my pull request.

Wilfred:
Wow, I'd assumed you were using 24-bit colour, but it's really impressive you've done this with the default terminal colours!

Here's before and after on one of the sample files:

Screenshot from 2022-05-23 23-03-18

Screenshot from 2022-05-23 23-01-26

Things that jump out at me:

* We probably don't need to make line numbers red/green with background highlighting
* It looks like subword bold highlighting (the `6` in `Version 0.6`) and delimiters (the `(` before `reverse`) is missing
* I don't think you need cansi, you should be able to calculate lengths before applying colours

egrieco:
I recently became the main organizer of a security meetup. I've been rather busy with the hand-off, hence the radio silence here.

> Things that jump out at me:
> 
>     * We probably don't need to make line numbers red/green with background highlighting
>     * It looks like subword bold highlighting (the `6` in `Version 0.6`) and delimiters (the `(` before `reverse`) is missing
>     * I don't think you need cansi, you should be able to calculate lengths before applying colours

Thanks for the guidance. I'll try to correct these issues soon. 😃

FranklinYu:
Would I be able to turn this off? My problem with 256-color is that they don’t work well when I SSH into another machine, because that machine won’t know the color scheme of frontend (SSH client). If I have multiple layers of SSH (not so rare in corp environment) it is much easier to handle the color scheme by the terminal itself, instead of telling every single CLI app to switch color. I ran from Delta simply because of this.

egrieco:
I have not forgotten about this issue, but enough time and changes have past that I will need to go back and ensure that everything is still working.

Also, I need to address the issues that @Wilfred raised:

* We probably don't need to make line numbers red/green with background highlighting
* It looks like subword bold highlighting (the `6` in `Version 0.6`) and delimiters (the `(` before `reverse`) is missing
* I don't think you need cansi, you should be able to calculate lengths before applying colours

egrieco:
Also, I totally agree that I shouldn't need to use `cansi` but I couldn't find where to grab the pre-styled text. Will take another look and bug you for help if I truly can't find the correct spot @Wilfred.

blag:
This looks amazing, I'm excited for it! Thank you!

tgrosinger:
It would be really wonderful to get this merged. I'm slightly colorblind and find that with just the thin text being colored it's very hard for me to tell if it's green or red. Adding a highlighted background would make a big difference.

Thank you!

vaxXxa:
[delta](https://github.com/dandavison/delta) also allows to emulate [diff-highlight and diff-so-fancy](https://dandavison.github.io/delta/diff-highlight-and-diff-so-fancy-emulation.html). For me `diff-so-fancy` highlighting style is more readable.

mattneub:
Ditto, please merge as there are entire lines of code that I cannot even see with the default colors.

muescha:
instead of fixed colors: should it make sense to add custom defined color themes so that styling can be adjusted by the user?

mattneub:
Or even without entire themes, wouldn't it be great if we could change individual colors as desired? Each color clearly corresponds to a particular "category" in difftastic's mind: "the name of a file", "code that was deleted", that sort of thing. Maybe it could be possible to show difftastic a config file that would set particular colors to something other then the default?

mattneub:
Thank you for thinking about this feature!

BanigaYBD (approved):

egrieco:
It's been quite a while since I created this pull request. Lots of updates have occurred since then, and as @Wilfred had pointed out previously, there are better ways to implement the coloring than I had cobbled together back then.

I'll need to clean it up or redo it entirely before it can be merged. I might get some time to do so in the next few weeks, depending on how much prep I need to do for a pending job offer.

egrieco:
Did a little playing around and got some basic background color and theme support working. There are still bugs so I'm not updating or replacing this pull request for now.

difftastic-screenshot-2024-03-25

If anyone wants to take a look, the current changes are here: https://github.com/egrieco/difftastic/tree/background_colors

muescha:
I like more the color scheme from the first post: changed lines in light green, changed words in dark green.

egrieco:
> I like more the color scheme from the first post: changed lines in light green, changed words in dark green.

Yeah, me too. This is just my attempt to redo the initial work on the new version, and to do so in a cleaner more maintainable way.

egrieco:
Just ran into a compilation error #689 while trying to rebase the new changes onto the latest HEAD.

I'll rebase onto the previous commit where things still compile...

Issue #293, opened 2022-05-27
Performance issue with the redundancy operations


yyzdtccjdtc:
https://github.com/Wilfred/difftastic/blob/f028b460037cd767adab832306cccfd3c4890705/src/display/hunks.rs#L579-L580

The struct LineNumber store a usize which is 8 byte. But under normal circumstances, it is very unlikely that a file will have more than 2^32 (4,294,967,295) lines of code. If we replace LineNumber with u32, we can reduce the redundant loading of 0's of the 4 high bytes then improve performance.

According to my tests, if we replace all the LineNumber within hunks.rs file with u32, the average execution time decreased from 4.2s to 3.8s, which is a 10% speedup.

Hope this information helps!

Wilfred:
Agreed, this seems like a worthwhile change.

Which files did you benchmark on? The diffing logic usually takes the majority of the time.

Wilfred:
I tried this: https://github.com/Wilfred/difftastic/commit/23f6a9f5c7cb904a23b8e3629ba7df4a3a62afbf and I only saw a ~0.3% improvement, measuring on elisp_before.el, load_before.js and Session_before.kt.

Would love a PR if you're doing something different :)

yyzdtccjdtc:
Thanks for your reply! These are the two cpp files I used for the benchmarking(Since github wouldn't let me upload .cpp format, I switched them to txt format.). I'll try to create a PR later for you to test.
[test2.txt](https://github.com/Wilfred/difftastic/files/8791612/test2.txt)
[test.txt](https://github.com/Wilfred/difftastic/files/8791611/test.txt)

Wilfred:
FWIW I also noticed that column numbers could be u32.

PR #297, opened 2022-05-30
Optimization for eliminating redundant memory operations

yyzdtccjdtc:
Optimization for eliminating redundant memory operations by changing LineNumber to u32.
https://github.com/Wilfred/difftastic/issues/293#issue-1251019903

At first I tried to make changes on the latest version, but I found that the latest version was running much slower than the 0.23.0 version that I initially used for testing and modifying (about 3m23s to 4s, which is about 50x slowdown). So now I only submit a pull request based on the 0.23.0 version (commit afb1b369f477cb3eaf1cee2a1e25cdf3d6fe2f9c)

These two attachments are the files I used for testing. Please let me know if you have any suggestions.

[test.cpp.txt](https://github.com/Wilfred/difftastic/files/8800785/test.cpp.txt)
[test2.cpp.txt](https://github.com/Wilfred/difftastic/files/8800786/test2.cpp.txt)


Wilfred:
Looks like the performance regression was between 0.28 and 0.29.

Wilfred:
Bisect points to 57a4b6d6ce9fbd367e557fccaf1ebf8975693c96, which will need analysing further.

PR #298, opened 2022-06-02
fix some typos

cuishuang:


Wilfred:
Thanks for the PR, but these are reference files that I don't want to change. It's easier to monitor difftastic performance if I don't change the sample files.

PR #301, opened 2022-06-26
use unicode-width to align CJK characters

### LINKED ISSUE #247: Calculate line length considering unicode
This is impossible to solve in the general case, but we could do better with the unicode-width crate.

lilydjwg:
This corrects this

difft-bad

into this

difft-good

It also fixes when line continuation number columns are colored when redirected to a file / pipe.

This closes #247.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -211,6 +211,7 @@ dependencies = [
  "tree-sitter",
  "tree_magic_mini",
  "typed-arena",
+ "unicode-width",
  "walkdir",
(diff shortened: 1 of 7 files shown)


lilydjwg:
I've also updated the sample_files and their expectations.

One of the other three "expected" updates is caused by a fullwidth emoji and others are removal of colored empty strings.

Wilfred:
Thanks! I've also confirmed that this doesn't affect performance :)

PR #305, opened 2022-07-01
feat: Add HTML parser

benmanns:
Adding an HTML parser based on [tree-sitter/tree-sitter-html](/tree-sitter/tree-sitter-html). Examples:

image

image

Some notes:

* It doesn't re-parse interior CSS or JS, so any change will diff as a full replacement.
* I configured sliders to be false. I don't really know what this does.
* Someone should maybe double check the `atom_nodes` and `delimiter_tokens`. I tried to match how some of the existing ones are configured, but I also don't really know what this does.
* I added a note about sliders.rs to adding_a_parser.md as I couldn't compile without configuring a match leg for the enum value. This could probably be expanded as (again... 😄) I don't know what it does exactly.


Wilfred:
Thanks! It'll be interesting to see how well this works :)

PR #315, opened 2022-07-13
Some format updates to README.md

TryWumbo:
can consider adding bullets to points under Known Issues and Non-goals for clarity and ease of update.


Wilfred:
Thanks for the PR! I think the current prose style reads a little better, but thanks for taking the time to submit this.

PR #353, opened 2022-09-07
Add arm64 releases for Linux and macOS

Niek:
This adds `aarch64` (arm64) releases for both Linux and macOS.


Wilfred:
Thanks!

Wilfred:
FWIW this isn't currently building on CI, so I've had to revert it. I've opened #368 as it would be nice to support.

PR #354, opened 2022-09-07
chinese translation of README.md

Yeegsing:
I add the chinese translation of README.md, and then add a Translation part on the original README.md.
All of the relative references work well.


Wilfred:
Wow! Thanks for doing this, this is really great :)

Issue #123, opened 2022-02-04
Add support for dynamically loading grammar libraries at runtime


jirutka:
Can you please add support for dynamically loading (any) grammar libraries at runtime? The current approach of vendoring every existing grammar and linking them in build-time is not scalable and not compatible with linux distributions.

This issue has been already solved e.g. in [diffsitter](https://github.com/afnanenayet/diffsitter) (https://github.com/afnanenayet/diffsitter/pull/177) or [helix](https://github.com/helix-editor/helix) (https://github.com/helix-editor/helix/pull/432).

You can either load a grammar library for the specified language from the system library path (this is typically `/lib:/usr/lib:/usr/local/lib`, but the path may vary between linux distros; the point is that you don’t depend on exact path, this is handled by libc) as any other system library – grammar libraries are expected to be named `libtree-sitter-<lang>.so` (diffsitter’s approach).

Or you can look up grammar libraries directly in some predefined directory (e.g. `/usr/lib/tree-sitter` or some app-specific dir) named as `<lang>.so` (helix’s approach).

The former is better, but both are acceptable for Linux distros – in the latter case, we can just symlink `<your-directory-for-grammars>` to e.g. `/usr/lib/tree-sitter` where system-wide grammars are installed by the system’s package manager (and used by all programs that use tree-sitter). These approaches are not mutually exclusive – you can support both.

Also, you can continue to provide some grammars with tree-grepper (as both diffsitter and helix do, they use git submodules) for the convenience of the users that build tree-grepper themselves and their distro doesn’t provide tree-sitter grammars yet. The aim is to _allow_ package maintainers to package grammars separately and give users the freedom to choose what grammars they will install.

Wilfred:
Ooh, are you interested in packaging difftastic @jirutka? That's be really neat.

This seems like a reasonable feature to add, but it'll require a few fiddly details:

(1) Difftastic doesn't just use the parser's C library, it uses the syntax highlighting file(s) (usually named `queries/highlights.scm`) too.

(2) For each language, difftastic also has configuration that specifies when to use the parser (e.g. file extensions) and how to use it (which AST nodes map to diffsitter's AST).

(3) Difftastic doesn't provide any configuration files yet, just a few CLI options and environment variables. Typically difftastic is invoked from git or hg rather than by the user, so I've generally preferred environment variables so far. I'm not sure what a good configuration system looks like for difftastic, so it'll need some design work.

[...]

PR #356, opened 2022-09-07
support loading parsers at runtime

mjyrkl:
only runtime loading with compile time configuration is implemented.
issue #123


hetky:
shared libraries are built with meson for coverage test

hetky:
rebased

hetky:
rebased

Wilfred:
Thanks for the PR! I agree this is a useful feature, but I can't merge it as-is:

(1) If it's using the system tree-sitter library, it needs to use the corresponding `highlights.scm` file. tree-sitter requires the names in the highlights library to match the parse tree.

(2) I don't understand why you've added meson. Is it necessary? Could it just use a Makefile instead?

(3) Could you add a test? I'd like to see the unit tests pass in this mode, or ideally `compare_all.sh`.

(4) Does `--list-languages` still work in this case? What happens if a user has `libtree-sitter-python` but not `libtree-sitter-ruby` installed?

(5) How do I add new languages in this model?

Wilfred (changes requested):
See comment above.

hetky:
(2) Replaced meson with cmake
(3) CI now runs unit tests and regression test for both static and dynamic linking
(4) `--list-languages` prints all supported languages but doesn't check if library is installed.
If user has libtree-sitter-python but not libtree-sitter-ruby installed:
```
# difft old.py new.py
--- normal output

# difft old.py new.rb
--- panic

# difft new.rb new.rb
--- panic

# difft --list-languages
--- normal output
```
(5) Cmake searches `parser.c,scanner.{c}c` in `tree-sitter-$lang-src` directories with hard coded handler for ocaml and typescript.

(5.1) If new language uses same naming scheme it will  be compiled automatically otherwise CI will probably fail. Special treatment is needed for some libraries (ocaml and typescript).

(5.2) Information about new language must be added in `dynlibs.rs` in `ts_function_name`, `ts_library_name` and `functions` mod.
In fact we can get name of function and library with `"tree-sitter-lang".replace("-","_)` and `format!("lib{tree-sitter-lang}")` if it's preferred.

Wilfred:
Thanks for following up! :)

Could we avoid panicking if a user tries to diff a file and the parser is absent? It's really frustrating to doing `git log` and difftastic crashes, as it stops showing commits beyond that point.

PR #359, opened 2022-09-08
Fix help-value for "--syntax-highlight" argument.

ingefredriksen:
"ON" and "OFF" in uppercase are not accepted. Print help in lowercase to match accepted values.


Wilfred:
Thanks :)

PR #360, opened 2022-09-08
Unset LC_ALL and LC_COLLATE to stabilize regression test output

yuja:
I set LC_COLLATE=C in ~/.profile, which appears to change the glob order. LC_ALL would also affect that, so let's unset both.


Wilfred:
Thanks, nice catch.

PR #361, opened 2022-09-08
Add support for QML

yuja:
QML is a UI language, and its syntax is basically JSON-like structure + JavaScript. The tree-sitter parser is named after the upstream grammar file qmljs.g, but the canonical language name is QML. So I choose Qml as the Language enum.
    
https://doc.qt.io/qt-6/qmlapplications.html


Wilfred:
Thanks!

PR #369, opened 2022-09-10
Add Hare support

esawady:
https://harelang.org


Wilfred:
This is looking pretty good! Could you fix the compare.expected file? It looks like it's reordered files, but I imagine this has been fixed by #360.

Wilfred (changes requested):
Once compare.expected is sorted, this is good to merge :)

esawady:
That didn't fix it for me (my guess is that the issue might be related to musl?) but I manually undid the reordering in compare.expected. ¯\\\_(ツ)\_/¯

Wilfred:
Weird. I should probably change the script to explicitly sort the file names rather than relying on order of files when globbing.

Wilfred:
Thanks for the PR! :)

PR #393, opened 2022-09-27
Some optimizations

QuarticCat:
First, enable thin-LTO. This brings ~5% speedup.

Before:

```text
 Performance counter stats for 'cargo run --release -- sample_files/slow_before.rs sample_files/slow_after.rs':

          1,033.66 msec task-clock:u              #    0.999 CPUs utilized          
                 0      context-switches:u        #    0.000 /sec                   
                 0      cpu-migrations:u          #    0.000 /sec                   
             6,127      page-faults:u             #    5.927 K/sec                  
     3,992,966,896      cycles:u                  #    3.863 GHz                    
       295,590,268      stalled-cycles-frontend:u #    7.40% frontend cycles idle   
     1,229,153,356      stalled-cycles-backend:u  #   30.78% backend cycles idle    
     4,948,279,348      instructions:u            #    1.24  insn per cycle         
                                                  #    0.25  stalled cycles per insn
       975,024,920      branches:u                #  943.275 M/sec                  
        20,612,498      branch-misses:u           #    2.11% of all branches        

       1.034353300 seconds time elapsed

       0.973504000 seconds user
       0.059968000 seconds sys
```

After:

```text
 Performance counter stats for 'cargo run --release -- sample_files/slow_before.rs sample_files/slow_after.rs':

            996.05 msec task-clock:u              #    0.998 CPUs utilized          
                 0      context-switches:u        #    0.000 /sec                   
                 0      cpu-migrations:u          #    0.000 /sec                   
             6,116      page-faults:u             #    6.140 K/sec                  
     3,926,308,401      cycles:u                  #    3.942 GHz                    
       263,648,312      stalled-cycles-frontend:u #    6.71% frontend cycles idle   
     1,349,401,185      stalled-cycles-backend:u  #   34.37% backend cycles idle    
     4,859,980,027      instructions:u            #    1.24  insn per cycle         
                                                  #    0.28  stalled cycles per insn
       920,523,681      branches:u                #  924.177 M/sec                  
        19,978,054      branch-misses:u           #    2.17% of all branches        

       0.998478582 seconds time elapsed

       0.932282000 seconds user
       0.063171000 seconds sys
```

The numbers of instructions are relatively stable.

I also measured them using hyperfine.

Before:

```console
$ hyperfine --warmup=3 'cargo run --release -- sample_files/slow_before.rs sample_files/slow_after.rs'
Benchmark 1: cargo run --release -- sample_files/slow_before.rs sample_files/slow_after.rs
[...]

PR #395, opened 2022-09-30
Some optimizations (cont)

QuarticCat:
This time I tried some radical optimizations.

### Benchmark Approach

To make the result more accurate, I updated my benchmark approach. Here's the command:

```zsh
cargo build --release &&
with-bench hyperfine --warmup=3 "$(echo ~/.cargo/target/release/difft sample_files/slow_{before,after}.rs)" &&
with-bench perf stat ~/.cargo/target/release/difft sample_files/slow_{before,after}.rs > /dev/null &&
time ~/.cargo/target/release/difft sample_files/slow_{before,after}.rs > /dev/null
```

where `with-bench` is a simple Zsh function that fixes CPU frequency and disables boost:

```zsh
_bench-start() {
    sudo cpupower frequency-set -u 3.6G -d 3.6G >/dev/null
    sudo sh -c 'echo 0 > /sys/devices/system/cpu/cpufreq/boost'
    echo '>>>>> BENCH START' >&2
}
_bench-end() {
    sudo cpupower frequency-set -u 10G -d 0.1G >/dev/null
    sudo sh -c 'echo 1 > /sys/devices/system/cpu/cpufreq/boost'
    echo '>>>>> BENCH END' >&2
}
with-bench() {
    _bench-start
    trap '_bench-end' EXIT INT
    $@
}
```

Note that this time `difft` is directly invoked instead of through `cargo run`, so the speedup percentage will be higher (`cargo run` has a fixed extra cost).

### Benchmark Results

Before my first PR:

```text
>>>>> BENCH START
Benchmark 1: /home/qc/.cargo/target/release/difft sample_files/slow_before.rs sample_files/slow_after.rs
  Time (mean ± σ):      1.120 s ±  0.032 s    [User: 1.080 s, System: 0.039 s]
  Range (min … max):    1.058 s …  1.169 s    10 runs
 
>>>>> BENCH END
>>>>> BENCH START

 Performance counter stats for '/home/qc/.cargo/target/release/difft sample_files/slow_before.rs sample_files/slow_after.rs':

          1,113.87 msec task-clock:u              #    0.998 CPUs utilized          
                 0      context-switches:u        #    0.000 /sec                   
                 0      cpu-migrations:u          #    0.000 /sec                   
             1,423      page-faults:u             #    1.278 K/sec                  
     3,844,769,039      cycles:u                  #    3.452 GHz                    
       319,285,519      stalled-cycles-frontend:u #    8.30% frontend cycles idle   
[...]

PR #398, opened 2022-10-03
Use successors

Masynchin:
Use successors instead of hand-made loop.

@QuarticCat I have slightly refactored your changes from #393, can you measure speed up/downgrading?

### DIFF
--- a/src/diff/stack.rs
+++ b/src/diff/stack.rs
@@ -1,3 +1,4 @@
+use std::iter::successors;
 use std::rc::Rc;
 
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


Masynchin:
@Wilfred it seems that tests doesn't cover `Stack::size`. For example,

~~~diff
pub fn size(&self) -> usize {
-    successors(self.head.as_ref(), |node| node.next.as_ref()).count()
+    successors(self.head.as_ref(), |node| node.next.as_ref()).count() + 100
}
~~~

passes the tests.

QuarticCat:
Actually, this function was only used in a debug output, which should not have an impact on the performance. And it was removed from #395.

QuarticCat:
By the way, I am planning the next big performance improvement PR, in which I completely refactored the stack. https://github.com/QuarticCat/difftastic/blob/new-thoughts/src/diff/graph.rs#L126

Masynchin:
> And it was removed in #395.

Seems that my PR is no longer relatable :(

Issue #376, opened 2022-09-18
Understand Python f-string syntax


ilius:
Right now I think difftastic treats Python's [f-string](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) as a normal string.
But it would be much nicer if it detected `{EXPR}` expressions inside it and treated them as normal Python expression (for example `f"a-b={a-b}"` or `f"{a-b=}"`), while also showing word-diff inside the f-string.

Issue #372, opened 2022-09-13
Better support for multiline strings


chshersh:
First of all, thanks a lot for creating this amazing tool! 🎉 
I use it daily and it helps me a lot 🤗 

While using the issue, I've noticed one potential improvement. Currently, when I edit a single line in a multiline YAML code block, I see the diff between two YAML values as the fully new string instead of detailed highlights in each string (as shown on the screenshot below)

![image](https://user-images.githubusercontent.com/4276606/189911704-67c48ea3-b188-42ba-a323-912e6a595f47.png)

I would be truly wonderful if difftastic could produce a more granular diff in this case! 🙏🏻 

Possibly related:

* https://github.com/Wilfred/difftastic/issues/178

**Additional information:**

+ OS: Ubuntu 22.04
+ Difftastic version: `0.35.0`

anka-213:
I wonder if it would be possible for difftastic to detect the nested language in these cases, to get full featured diffs even inside strings? But even a simple linewise diff would be an improvement of the current mode of treating strings as atomic.

Ramblurr:
With the merging of https://github.com/Wilfred/difftastic/commit/9133918dd40d7e55e605934994b87807728b201d this should be possible now right?

I deal with lots of yaml  with long multiline strings, and this feature would be really really helpful:

![image](https://github.com/Wilfred/difftastic/assets/14830/95eb8ded-14d8-46a3-98e2-a88e4a6e8e6f)

karlb:
Multiline yaml diffs in one of the rare cases where a normal diff is more readable than difftastic's output. Here's one example:

<img width="2497" height="554" alt="Image" src="https://github.com/user-attachments/assets/5efd1028-70ee-4fcd-bb39-cd013135ad30" />

I see two problems that might be related:
* The unchanged lines are included in the red/green diff
* The bold part with the actual change goes over two lines, while the semantic change is only a single line

So in this case a line-based diff of the multiline part is preferable.

Tested with
```
Difftastic 0.65.0
Toolchain: 1.90.0
System:    macos aarch64
```

Thanks for difftastic, it is an excellent tool!

Issue #178, opened 2022-03-24
Highlight the changed lines in a multiline string more prominently


Wilfred:

mortenfyhn:
This would be great. I love difftastic, but it's inconvenient to not easily see what's changed between two long strings. I notice this in Python and C++, but I suppose the handling is similar in other languages. Here's an example:

![image](https://user-images.githubusercontent.com/6639836/165501327-76568979-41af-4ac4-b62d-fbfde7610587.png)

(The difference is the `non` on the last line.)

Somehow highlighting the changed words within strings would be great. I see the same behaviour in C++ with just a long single-line string.

vwkd:
Could this be solved by an option to customize the line style (green / red text color) differently from the change style (bold green / red text color)?

For example, one could use a dimmer background color instead of a text color for the line style, which would make the changes pop out more.

Possibly related https://github.com/Wilfred/difftastic/issues/167

Issue #32, opened 2021-09-06
Syntax highlight in the diff output


vlsi:
It would be nice if there was syntax highlighting in the diff output, so the code structure is easier to understand.

Sample PL/SQL diff:

git version 2.33.0
<img width="1057" alt="diff with git 2.33" src="https://user-images.githubusercontent.com/213894/132167666-c92a0f44-a692-40e3-9c3b-b7a3a4c48d84.png">

difftastic 0.8
<img width="1413" alt="diff with difftastic 0.8" src="https://user-images.githubusercontent.com/213894/132167697-b93cfde1-e2ce-4c5e-b9b9-c55e4e220433.png">

delta 0.8.3
<img width="1123" alt="diff with delta 0.8.3" src="https://user-images.githubusercontent.com/213894/132167724-41acc709-910c-4a46-8b3a-4303bbd41c56.png">

---

PL/SQL diff where the diff is within string literals (the indentation at line 256 looks odd, and I did not notice it before I took the screenshots).
The code is "PL/SQL code generator written in PL/SQL", and I do not expect that PL/SQL code within string literals to be highlighted as PL/SQL :) However, it would be nice if string literals were highlighted differently from the regular code.

git version 2.33.0
<img width="1010" alt="git version 2.33.0" src="https://user-images.githubusercontent.com/213894/132168336-1087393c-a9b5-426e-94fc-a93d9a0ef54f.png">

difftastic 0.8
<img width="1410" alt="difftastic 0.8" src="https://user-images.githubusercontent.com/213894/132168354-992084f2-fdec-48c2-b1ef-a36eda8e19dc.png">

delta 0.8.3
<img width="1073" alt="delta 0.8.3" src="https://user-images.githubusercontent.com/213894/132168368-ac2a848f-571a-46ae-8ea0-94e6e1e232b2.png">

---

Re "string literals vs code" there's a semi-standard extension from IntelliJ: https://www.jetbrains.com/help/idea/using-language-injections.html#use-language-injection-comments

For instance, the following bit in Groovy code (Jenkins Job DSL code) is highlighted as XML in IntelliJ IDEA:

```
// $/ ... /$ is a multi-line string in Groovy
//  language=xml comment triggers "inject XML into the following literal" in IDEA
+ echo('> config/config.xml', /* language=xml */ $/
<root>
  <profile>
    <datasources>
    </datasources>
  </profile>
</root>
/$) +
```

Wilfred:
[...]

PR #409, opened 2022-10-18
Support parsing of sub-languages

### LINKED ISSUE #382: Diffing a modified HTML script element highlights its whole content
This is almost surely a feature request, or perhaps just an observation, rather than a bug.

> (1) A description of the issue. A screenshot is often helpful too.

If i have an HTML document with a script element containing JavaScript code, and i modify the JavaScript code, then difftastic's diff highlights the whole content of the script element as changed. The same goes for a style element containing CSS which i modify.

I would guess this is because the parser for HTML is in effect, and doesn't attempt to parse the content of the script element.

In my dream diff tool, the parser would parse the content of the script element with a JavaScript parser. And would similarly switch language when doing things like parsing JavaScript string constants containing HTML, or bash here-docs containing YAML, and so on and so forth. This will probably remain a dream!

A modest but significant improvement here might be to see that the content of the script element is multiple lines, and apply a fallback line-by line diff.

> (2) A copy of what you're diffing. If you're diffing files, include the before and after files. If you're using difftastic with a VCS repository (e.g. git), include the URL and commit hash.

This is a private repository of my employer's, so i can't share it. If a concrete example would be useful, i am happy to construct one.

> (3) The version of difftastic you're using (see `difft --version`) and your operating system.

```
$ difft --version
Difftastic 0.36.1
$ grep PRETTY_NAME /etc/os-release
PRETTY_NAME="Ubuntu 20.04.5 LTS"
```

sesse:
This allows given nodes (configurable per-language, using tree-sitter's query syntax) to be re-parsed as other languages. The canonical example is CSS or JavaScript inside HTML, which normally would be a single token but now can get the full range of syntax highlighting and tree diffing.

The config sets this up for only two languages: HTML (contains CSS or JavaScript in <script> or <style> tags; we don't support style="" or onclick="" etc. at this point), and Makefiles (contains Bash in $(shell ...) commands). The latter is fairly obscure; the big win is in the former.

It would be nice to also have this support for PHP; however, the HTML parser seems to be a bit confused when asked to parse the partial HTML blocks we get if we just mark the “text” blocks as HTML, so for this to work well, probably the PHP blocks should be parsed as sub-languages of HTML instead of vice versa.

Also, as a minor quibble, there should be support for bash in Perl's backticks (similar to in Makefiles), but the tree-sitter Perl parser does not support backticks at all (it goes into error recovery).

There may have been languages that I've missed, e.g. some languages might have nodes that contain e.g. SQL.

Fixes #382. Potentially relevant to #376.

### DIFF
--- a/sample_files/compare.expected
+++ b/sample_files/compare.expected
@@ -62,7 +62,7 @@ sample_files/helpful-unit-test-before.el sample_files/helpful-unit-test-after.el
 79597af48ff80bcf9f5d02d20c51606d  -
 
 sample_files/html_before.html sample_files/html_after.html
-5bf6a3d864706e18932c8cd027cb4c0b  -
+baf2ef8bffad85273b97d7ae41d67e0e  -
 
 sample_files/html_simple_before.html sample_files/html_simple_after.html
 2c5a14df5b793bc136e37f263733b26f  -
--- a/src/parse/tree_sitter_parser.rs
+++ b/src/parse/tree_sitter_parser.rs
@@ -1,5 +1,6 @@
 //! Load and configure parsers written with tree-sitter.
 
+use std::collections::HashMap;
 use std::collections::HashSet;
 
 use crate::parse::guess_language as guess;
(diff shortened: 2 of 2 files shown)


anka-213:
Could also be useful for #372 and https://github.com/Wilfred/difftastic/issues/32#issuecomment-1290588760 (and probably other instances of #178), since those are both examples of bash code being nested inside a string in a different language. The difficult part there is how to tell if it's bash code that's nested or something else, since both `nix` and `yaml` allows putting anything in the multiline strings, it's just that bash is common.

It is also possible that the easiest solution for those would be a basic fix of #178, e.g. by treating multiline strings as a sequence of lines instead of as an atomic blob.

Wilfred:
Merged in 9133918dd, thank you :)

Wilfred:
@sesse FWIW I found a bug with this, which you might want to cherry-pick to your fork: 63a3bf0c915a49fb18324a25cc4c4be7334c6f72

PR #414, opened 2022-10-26
Skip ignored nodes when converting TS nodes to difftastic

FoamScience:
The objective of this PR is to implement a feature for **ignoring specific  Tree Sitter nodes** and skip converting them to difftastic nodes.

As a I'm a "wanabe a Rust dev",  and this is my first contribution to someone else's code, the first commit takes it easy;
only adds an extra setting in `TreeSitterConfig`.

With `file1.cpp`:
```cpp
int main (int argc, char *argv[])
{
    // This is main
    // And this is an extra comment
    return 0;
}
```
and `file2.cpp`:
```cpp
int main (int argc, char *argv[])
{
    // This is also main, but a bit different
    const int i = 0;
    return 0;
}
```
The diffs come out as (effectively ignoring changes in comments, with inline display):

While I think it's pretty neat,  I see two issues I probably need help with:
1. The default display mode still displays comment additions/removals. Need to get rid of this too. But I guess some might consider it a feature:

2. Instead of hard-coding nodes to be ignored, we obviously need a way to specify them at run-time, probably through CLI options.

The issue is that I'm fairly new to Rust and don't know how to cleanly read a collection from CLI. Maybe [this](https://github.com/snowpoke/collections-fromstr)?
If I can get some help, or at least orientation, It would be much appreciated!

> In this commit, I turned on comments ignoring only for C++ as I plan to select what to ignore at run-time.
> If you like the idea and want to merge this feature, you'll have to teach me how to read in these lists 😄, or switch to a config file

### DIFF
--- a/src/parse/tree_sitter_parser.rs
+++ b/src/parse/tree_sitter_parser.rs
@@ -37,6 +37,10 @@ pub struct TreeSitterConfig {
     /// Tree-sitter query used for syntax highlighting this
     /// language.
     highlight_query: ts::Query,
+
+    /// Tree-sitter nodes that we ignore and do not convert to
+    /// difftatic nodes
+    ignored_nodes: HashSet<&'static str>,
 }
 
 extern "C" {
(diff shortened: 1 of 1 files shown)


FoamScience:
Also, I guess this opens the way to maybe support ignoring full S-expression queries from Tree Sitter in the future.

Wilfred:
Hi @FoamScience! Thanks for the PR!

I ended up building `--ignore-comments` in #449, so you're welcome to look at the implementation details I used if you're interested.

Your pull request was a good start, best of luck on your rust journey :)

Issue #424, opened 2022-11-15
Cargo build on windows is broken


CGQAQ:
```ps1
cargo b
  Downloaded bytecount v0.6.2
  Downloaded indexmap v1.7.0
  Downloaded os_str_bytes v6.0.1
  Downloaded crossbeam-epoch v0.9.8
  Downloaded crossbeam-utils v0.8.8
  Downloaded quote v1.0.18
  Downloaded clap v3.1.18
  Downloaded const_format_proc_macros v0.2.22
  Downloaded once_cell v1.12.0
  Downloaded regex-syntax v0.6.26
  Downloaded unicode-xid v0.2.3
  Downloaded crossbeam-channel v0.5.4
  Downloaded owo-colors v3.4.0
  Downloaded clap_lex v0.2.0
  Downloaded regex v1.5.6
  Downloaded unicode-ident v1.0.0
  Downloaded const_format v0.2.23
  Downloaded proc-macro2 v1.0.39
  Downloaded terminal_size v0.2.1
  Downloaded petgraph v0.6.1
  Downloaded 20 crates (1.5 MB) in 5.23s
   Compiling autocfg v1.1.0
   Compiling winapi v0.3.9
   Compiling cfg-if v1.0.0
   Compiling lazy_static v1.4.0
   Compiling cc v1.0.73
   Compiling crossbeam-utils v0.8.8
   Compiling memchr v2.5.0
   Compiling scopeguard v1.1.0
   Compiling rayon-core v1.9.3
   Compiling proc-macro2 v1.0.39
   Compiling regex-syntax v0.6.26
   Compiling log v0.4.17
   Compiling windows_x86_64_msvc v0.36.1
   Compiling either v1.6.1
   Compiling unicode-ident v1.0.0
   Compiling hashbrown v0.11.2
   Compiling num_cpus v1.13.1
   Compiling quick-error v1.2.3
   Compiling minimal-lexical v0.2.1
   Compiling os_str_bytes v6.0.1
   Compiling unicode-xid v0.2.3
   Compiling fixedbitset v0.4.1
   Compiling libc v0.2.126
   Compiling version_check v0.9.4
   Compiling fnv v1.0.7
   Compiling bytecount v0.6.2
   Compiling once_cell v1.12.0
   Compiling bitflags v1.3.2
   Compiling strsim v0.10.0
   Compiling rustc-hash v1.1.0
   Compiling radix-heap v0.4.2
   Compiling typed-arena v2.0.1
[...]

Issue #67, opened 2021-12-05
cargo install difftastic fails from a standard windows console with VS2022 installed


jessesna:
`cargo install difftastic` fails from a standard windows console with VS2022 installed.
It works from within a VS2019 developer prompt.

Here's the output from the windows console:
```
error: linking with `link.exe` failed: exit code: 1120
  |
[...]

PR #425, opened 2022-11-15
don't use softlink for cross platform building

CGQAQ:
FIXED: #426
FIXED: #424
FIXED: #67


CGQAQ:
image

it should not fail 'cause `vendor/tree-sitter-json/src/parser.c` exists

CGQAQ:
@Wilfred could you please take a look

Wilfred:
If you can get the build to go green, I'd be happy to remove the symlinks. They are currently necessary though. Cargo will refuse to package directories that contain `Cargo.toml` files, so I need a symlink to the `vendor/foo/src` directory to ensure everything goes in the package. It's also important for package size: difftastic is already over the 10MiB default limit for package size, even after my changes.

Kreijstal:
Why not have subpackages?

Wilfred:
They're C libraries, not crates FWIW.

PR #434, opened 2022-11-27
Modify. For the second and subsequent hunks, difft outputs file name in the default color.

### LINKED ISSUE #400: Only list files once, not for each hunk
First off, thank you for this project, it's wonderful!

Second, I apologize for the terrible image editing I did to illustrate my point here. They are terrible. Thank you for persevering through them.

That being said, I think there's a low-hanging fruit type of improvement that can be made to the display when displaying changes in multiple files.

The yellow color of the filename, and its large contrast with the black background, makes these changes all look like they happened in different files:

Screen Shot 2022-10-04 at 2 58 20 PM

The only thing that indicates that they are different hunks in the same files is the `X / Y` identifier after the filename. It is difficult to jump to the next file, because I have to visually read and cognitively compare the filenames and hunk numbers with the one above it as I go down the page. My eyes and brain have to do this motion:

Screen Shot 2022-10-04 at 2 58 20 PM copy 2

I think it would be better to use color and contrast to differentiate hunks in different files, and less contrast to indicate hunks within the same file, like so:

Screen Shot 2022-10-04 at 2 58 20 PM copy

I didn't show it, but the next file would have the its filename in yellow, and the same grayish hunk identifier/s.

So really the only change that needs to be made is omitting displaying the filename for non-first hunks. I hope this is easy to do.

I think this would make it much easier to quickly scan through the output for different files, as the high-contrast yellow would differentiate changes between separate files a bit better than it does now. My eyes can jump from one yellow/high-contrast filename to the next instead:

Screen Shot 2022-10-04 at 2 58 20 PM copy

(I forgot to edit in an empty line between the second filename and the preceding content, but let's just pretend it's there.)

There's a lot less to visually and cognitively process, so it should be easier and faster to quickly scan through.

Difftastic 0.27.0 on macOS 11.6.8 Big Sur (Intel), installed via Homebrew IIRC.

rhirano0715:
Resolves #400

For the second and subsequent hanks, difft outputs file name in the default color.
I thought it would be inconvenient to hide file names when there are many hunks, so I changed the output to use the default color.

Output Example:

OutputExample

### DIFF
--- a/sample_files/dir_after/has_many_hunk.py
+++ b/sample_files/dir_after/has_many_hunk.py
@@ -1,0 +1,404 @@
+def function001():
+    print("function001 start")
+
+def function002():
+    pass
+
+def function003():
+    pass
+
+def function004():
+    pass
+
+def function005():
+    pass
+
+def function006():
+    pass
+
+def function007():
+    pass
+
+def function008():
+    pass
+
+def function009():
+    pass
+
+def function010():
+    pass
+
+def function011():
+    print("function011 start")
+
+def function012():
+    pass
+
+def function013():
+    pass
+
+def function014():
+    pass
+
+def function015():
+    pass
+
+def function016():
+    pass
+
+def function017():
+    pass
+
+def function018():
+    pass
+
+def function019():
+    pass
+
+def function020():
+    pass
+
+def function021():
+    print("function021 start")
+    pass
+
+def function022():
+    pass
+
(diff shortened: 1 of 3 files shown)


Wilfred:
Merged in 436edb2ab48e4a5ae76ae06de2f4ca3de1dcbbcb, thank you :)

PR #440, opened 2022-12-07
Add default-parser option

### LINKED ISSUE #439: Allowing for user to choose default text parsing
Sometimes difftastic will parse a file incorrectly either due to an error in the parser, an error in the file itself or a complicated situation not yet handled by the parsers (e.g. helm templates, which are kind-of yaml files with embedded go-template, see #267 and #377). In those cases it would be helpful if the user could force difftastic to use the default text parsing instead of showing "No syntatic changes."

I envision something like: `difft --language default` or `difft --default-parser`

I'm willing to make a PR but I don't know rust and don't know where should I start 🙈

(1) One idea would be to treat the default text parser the same as the other parsers. I would add a `default` language in the [languages enum](https://github.com/Wilfred/difftastic/blob/master/src/parse/guess_language.rs#L20) and then instead of matching to `None` we would match to this `default`. For example here:
- https://github.com/Wilfred/difftastic/blob/master/src/options.rs#L277
- https://github.com/Wilfred/difftastic/blob/master/src/main.rs#L385

But that would change the design choice of using `None` to fallback to text parsing. Another alternative to this idea would be to re-design the whole `line_parser` to fit the vendor parsers format. I have absolutely no idea how I would do that 😅 

(2) Another idea would be adding a `--default-parser` flag and then ignore `language_override`. For example [here](https://github.com/Wilfred/difftastic/blob/master/src/main.rs#L299): 
```rust
let language = language_override.or_else(|| guess(guess_path, guess_src));
```
would become something like:
```rust
let language = if default_parser {
  None;
} else {
  language_override.or_else(|| guess(path, &src));
}
```

The important point here is side-stepping the `guess` function.

I tend to prefer Option (2) because it seems less intrusive to me, but it still adds another top level flag which could maybe be covered by an already existing flag. I'll try to implement that one but am very welcome to feedback, specially if I'm tackling this issue completely wrong!

mparada:
The flag `--default-parser` forces the use of the default and fallback line_parser.

Fixes #439


Wilfred:
Thanks for the PR! I ended up going with an option that allows you to control language detection based on globs.

PR #467, opened 2023-01-19
Update linguist-vendored attribute

sgkoishi:
https://github.com/github/linguist/blob/master/docs/overrides.md#vendored-code
Use `**` instead of `*` to exclude nested files.


Wilfred:
Thanks :)

Issue #433, opened 2022-11-24
difft@0.38.0 misclassifying JS files as binary. Sometimes. Depending on file size(?)


robertklep:
I recently upgraded to 0.38.0 from (I think) 0.35, using Homebrew on macOS Ventura 13.0 (M1 hardware), and it started to misclassify some of my Javascript files as binary.

I was able to narrow it down to a test with these two files:
(one.js)
```
if (! window.Home) {
  window.Home = new (class Home {
    constructor() {
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
      // random
    }
  })();
}
```
(two.js)
```
```
(yes, an empty file)

This yields the following result:
```
$ difft one.js two.js
two.js --- binary
Binary contents changed.
```

If I remove one of the `// random` lines, it works as expected.

Wilfred:
Looks like the MIME type database is categorising the JS file as audio/x-mod. Difftastic should probably treat pure-ASCII files as text unconditionally.

robertklep:
Ah, that would explain the issue 😊 JS files won't always be pure-ASCII though.

PR #473, opened 2023-01-31
Fix file content type detection for small gzipped files

Miksu82:
This PR fixes a case where small gzipped files are incorrectly interpreted is UTF-16 strings. See the added test case for explanation. I encountered this issue in one of my employers private repos.

I admit the fix isn't pretty. Fixing #433 seems to have made the content type detection too complicated. I tried to reproduce #433 in 2e7c90c472fbfe9df93d2c0e63817ba7f8491deb but I wasn't able to do that. However I decided to be on the safe side and not revert those changes.

In the long run it would probably be beneficial to consider all supported file types (as defined in `guess_language.rs`) to be plain text and for other files trust the content type check from `tree_magic`. 

I can create a new PR to implement that if the maintainers think it is a good idea.


Miksu82:
Hmm... the tests fail. But they do not fail on my machine. I'll try to see what is happening there.

I'll also fix the email address in the commits so that they point to the correct user

Miksu82:
I was able to reproduce the test failure when running the tests in https://hub.docker.com/_/rust

Should be fixed now.

Wilfred:
Yeah, it's surprising how many binary files are valid UTF-8 or UTF-16. Coming up with good heuristics is hard.

FWIW I suspect the reason you're seeing test differences is that the magic_mini repo relies on the local copy of the file database, which CI probably doesn't have.

src/files.rs
```diff
@@ -137,25 +137,35 @@ pub fn guess_content(bytes: &[u8]) -> ProbableFileKind {
...
     }
 
+    let mime = tree_magic_mini::from_u8(magic_bytes);
+    info!("MIME type detected: {}", mime);
+
+    // This is a dirty hack to handle cases where files are obviously
+    // binary but can be still converted to UTF16 string.
+    // See test_gzip_is_binary
+    match mime {
+        "application/octet-stream" => return ProbableFileKind::Binary,
```
Wilfred, line 154: This isn't quite right.

```
$ RUST_LOG=info cargo run -- sample_files/utf16_before.py sample_files/utf16_after.py
 2023-02-01T23:39:05.289Z INFO  difft::options > CLI arguments: ["sample_files/utf16_before.py", "sample_files/utf16_after.py"]
 2023-02-01T23:39:05.305Z INFO  difft::files   > MIME type detected: application/octet-stream
 2023-02-01T23:39:05.307Z INFO  difft::files   > MIME type detected: application/octet-stream
```

The MIME database treats valid UTF-16 files as octet-stream.

```suggestion
```

Change this check to only consider gzip files, and I'm happy :)

(If you can see better heuristics that would be wonderful, but nothing is obvious to me at least.)

Wilfred (changes requested):
This has broken UTF-16, see my comment for a suggested fix :)

src/files.rs
Miksu82: Done. You were right that the differences in my local runs and CI was that I didn't have the mime database so tree_magic interpreted pretty much everything to have `application/octet-stream` content type. I installed and setup the db properly and I was able to match the same results as the CI

Wilfred:
Merged, thank you! :)

Issue #296, opened 2022-05-29
`git log` and `git show` fail on Windows with "No such file: /dev/null"


cgranade:
When using `git show <commit-id> --ext-diff` on Windows for commits that include creating a new file, an error is raised by `difft` indicating that it was called with `/dev/null` as an argument, despite that no such file exists on Windows:

```
git show HEAD --ext-diff
commit a77b7f947a73da887d2cc7692b17372d115a7961 (HEAD -> main, origin/main, origin/HEAD)
Author: ...
Date:   ...

No such file: /dev/null
fatal: external diff died, stopping at ...
```

- Difftastic 0.28.0
- Windows 11 22610.1

Wilfred:
Interesting, thanks. I'm guessing that git calls the diff tool with `/dev/null` for new files, regardless of whether it's on a platform that has `/dev/null` as a valid path.

nzbart:
@Wilfred I'm still experiencing this in `Difftastic 0.37.0`.

OndrejNepozitek:
Also happens to me, Windows 10, Difftastic 0.41.0. Unfortunately, this makes it impossible to use difftastic with git on Windows.

Wilfred:
@OndrejNepozitek could you try setting `RUST_LOG=info` and running git again?

This should show what arguments difftastic is getting, e.g.

```
2023-01-15T23:13:10.838Z INFO  difft::options > CLI arguments: ["src/display/hunks.rs", "/tmp/git-blob-Lb98A6/hunks.rs", "08e25505d8b9373c6265a99a5b775e2dc4f4bf6c", "100644",
```

This should hopefully show the `/dev/null` string somewhere, but I'm guessing it's slightly different on Windows.

n9v9:
Hi, I found this issue yesterday and decided to try and fix it.

Example to reproduce this issue:

```
mkdir repo
cd repo
git init
git config diff.external difft
copy nul foo.txt
git add foo.txt
git diff --staged
```

Enable `RUST_LOG`:

```
set RUST_LOG=info
```

Output of `git diff`:

```
2023-02-01T16:26:03.783Z INFO  difft::options > CLI arguments: ["foo.txt", "/dev/null", ".", ".", "foo.txt", "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391", "100644"]
No such file: /dev/null
fatal: external diff died, stopping at foo.txt
```

If committing this change and then deleting the file with:

```
git commit -m Initial
[...]

PR #475, opened 2023-02-01
Check arguments passed from Git for `/dev/null`

n9v9:
This fixes https://github.com/Wilfred/difftastic/issues/296.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -295,9 +295,14 @@ impl FileArgument {
     }
 
     /// Return a `FileArgument` that always represents a path that
-    /// exists.
+    /// exists, with the exception of `/dev/null`, which is turned into [FileArgument::DevNull].
     pub fn from_path_argument(arg: &OsStr) -> Self {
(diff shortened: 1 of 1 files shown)


Wilfred (changes requested):
This isn't quite right -- `from_path_argument` is intended for paths that always exist.

I suggest you replace the calls to `from_path_argument` to `from_cli_argument` here:

https://github.com/Wilfred/difftastic/blob/master/src/options.rs#L412-L433

n9v9:
Thanks for the feedback. I changed `from_path_argument` because `from_cli_argument` additionally checks for `-` and uses stdin for it. But when a file called `-` exists (which hopefully shouldn't ^^), then git passes `-` as the file name and `from_cli_argument` would block waiting to receive data on stdin.

Should I still change it to use `from_cli_argument` or something else?

Wilfred:
Oh, that's a really great point. Thanks, merged :)

PR #483, opened 2023-02-12
Add racket support

6cdh:


6cdh:
I think it works now.
I also found a problem. If `here_string` not in `atom_nodes`,
It will show a weird result:

Screenshot_20230212_140457

What happened here? Does it considered a parser error?

6cdh:
I just fixed another bug in `here_string`. It's done now.
The problem above still exists.

Wilfred:
This is a limitation of difftastic today: it doesn't handle string interpolation well. Difftastic assumes that all parts of the syntax tree have their own position, whereas things like `"foo $bar baz"` tend to just have `(string_lit outer_position (interpolated inner_position '$baz'))`. Treating string literals like heredocs as atoms is the right thing to do.

Could you add syntax highlighting too? I see https://github.com/6cdh/tree-sitter-racket/blob/main/queries/highlights.scm exists.

6cdh:
It's completed now.

Wilfred:
Merged in 5e659e2d9862d10499850a762212fcd94ba84181, thanks!

PR #485, opened 2023-02-16
Add support for Newick files

delehef:
[Newick files](https://en.wikipedia.org/wiki/Newick_format) are very commonly used in bioinformatics to describe tree-like structures, typically phylogenies.

This PR add support for Newick files to Difftastic. I understand that it may be out of scope for this project, so please don't hesitate to reject it if that's the case.


Wilfred:
I know nothing about bioinformatics. It sounds like these files are structured text files for trees? That does sound like a good fit for difftastic.

Have you tried it on some real files? How well did it work?

I'm not opposed to adding this, but I think it needs a sample file in `sample_files/` to make sure the results are sensible.

delehef:
> It sounds like these files are structured text files for trees?

Exactly. They are basically S-exp-with-commas based tree representations with a few biology-specific metadata tossed in.

> Have you tried it on some real files? How well did it work?

Yes, I have been using it myself for a few weeks now. Diffs can get heavy to read on large files, but I still find it pretty useful

> I think it needs a sample file in sample_files/ to make sure the results are sensible.

Sure, I added two files representing the evolution of the species stored in different versions of the [ENSEMBL genomic database](https://en.wikipedia.org/wiki/Ensembl_genome_database_project).

Wilfred:
Merged in 9556cd978, thanks :)

PR #486, opened 2023-02-16
add option and help text for --strip-trailing-cr feature

spital:
This is a work in progress, no functionality present at the moment.
What are the thoughts please, is it worth developing ?

image

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -249,6 +249,10 @@ fn app() -> clap::Command<'static> {
                 .hide(true)
                 .allow_invalid_utf8(true),
         )
+        .arg(
+            Arg::new("strip-trailing-cr").long("strip-trailing-cr")
+                .help("Strip carriage return on input (CRLF -> LF).")
+        )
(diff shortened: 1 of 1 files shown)


Wilfred:
What's the goal here? Are you trying to normalise the output, or change the input to the diffing logic?

spital:
I should have been more verbose. It is the same as with normal diff. As the name says, it strips the trailing CR in input files. If you have one file created on Windows (CRLF end of lines) and you diff it against one on linux e.g. processed with dos2unix (LF endings), you get that the whole file was changed, which is wrong, files are the same, only line endings are not.

Narcha:
This would be very useful. I use WSL at work, and difftastic is almost useless for diffing files across the WSL boundary because of different line endings. I have to resort to `diff --ignore-space-change` in these cases, but I would much rather use difftastic.

Wilfred:
Done, thanks for the examples! :)

Narcha:
Thank you :heart:

PR #494, opened 2023-03-13
Add Ada support

karlding:


karlding:
Sorry, I noticed some typos in the commit messages so I've force-pushed to my branch. The diff otherwise remains unchanged.

src/parse/guess_language.rs
```diff
@@ -224,6 +227,7 @@ pub const LANG_EXTENSIONS: &[(Language, &[&str])] = &[
 
 /// Which file names are associated with which languages.
 pub const LANG_FILE_NAMES: &[(Language, &[&str])] = &[
+    (Ada, &["alire.toml"]),
```
Wilfred, line 230:
```suggestion
```
I would expect `.toml` files to be parsed  as TOML, so I don't think this is what you want.

Other than that, this looks great!

karlding: Ah yes, that's what I get for not actually reading what the code uses `LANG_FILE_NAMES` for.

For some reason I had the impression that this was doing something more sophisticated---given some project folder, it would magically use the presence of certain files to guess the language of other files.

PR #502, opened 2023-04-05
Add Solidity support

vquelque:
This PR adds Solidity support to difftastic, thanks to [tree-sitter-solidity](https://github.com/JoranHonig/tree-sitter-solidity)


Wilfred:
Thanks! I've also updated https://github.com/JoranHonig/tree-sitter-solidity/issues/36 with a repro after playing with this.

PR #507, opened 2023-04-28
doc: extend description of mercurial configuration

grunweg:
Mention that command-line arguments can be added, and that `hg dft` supports all arguments `hg diff` does.
Seeing the `hg dft` incantation below, it took me a moment to realise that `hg dft` will then support comparing revisions.

Perhaps a better change would be to tweak the description of the `hg log -p` section: any two revisions can be compared like this, not just the working directory to the latest revision checked out.

### DIFF
--- a/manual/src/mercurial.md
+++ b/manual/src/mercurial.md
@@ -10,8 +10,8 @@ your `.hgrc`.
 extdiff =
 ```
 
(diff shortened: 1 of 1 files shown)


Wilfred:
Thanks, that's definitely helpful info :)

PR #511, opened 2023-05-05
Add LaTeX parser

grunweg:
Based on https://github.com/latex-lsp/tree-sitter-latex.
 ~~Not tested yet~~Needs a regression test. Feel free to jump on this and help!


grunweg:
I see some failing builds. I'm at a loss at the precise reason, help appreciated. (All changes I made are in the last three commits.)

build.rs
```diff
@@ -230,6 +230,11 @@ fn main() {
             src_dir: "vendored_parsers/tree-sitter-kotlin-src",
             extra_files: vec!["scanner.c"],
         },
+        TreeSitterParser {
+            name: "tree-sitter-latex",
+            src_dir: "vendored_parsers/tree-sitter-latex-src",
+            extra_files: vec![],
```
Wilfred, line 236: It looks like you need to add some additional files here, as the compiler can't find all the dependencies of the LaTeX parser.

grunweg:
Small update: I've fixed language detection by correction the file extensions, [matching upstream](https://github.com/latex-lsp/tree-sitter-latex/blob/master/package.json) - and squashed commits directly.

build.rs
grunweg: Looking [upstream](https://github.com/latex-lsp/tree-sitter-latex/blob/master/binding.gyp), it seems that `parser.c` should also be included. However, doing so yields a linker error about multiple definitions.

```note: /usr/bin/ld: [redacted]/difftastic/target/debug/build/difftastic-e8d4711a8939d520/out/libtree-sitter-latex.a(parser.o): in function `tree_sitter_latex':
          [redacted]/difftastic/vendored_parsers/tree-sitter-latex-src/parser.c:651664: multiple definition of `tree_sitter_latex'; [redacted]/difftastic/target/debug/build/difftastic-e8d4711a8939d520/out/libtree-sitter-latex.a(parser.o):[redacted]/difftastic/vendored_parsers/tree-sitter-latex-src/parser.c:651664: first defined here```

Indeed, `parser.c` on line 651664 defines the function tree_sitter_latex, but I cannot find other definitions in C. (Only for the bindings to node, swift and Rust - but IIUC, those are required.)

grunweg:
Tested with https://github.com/florian-rabe/Teaching/ and https://github.com/ProofSystem/Encyclopedia; output looks reasonable.
From my side, this is ready for review; it's just missing a regression test.

Wilfred:
Merged, thanks :)

Seems to work pretty well, but I haven't tested it much.

PR #513, opened 2023-05-10
Document how to use difftastic with Fossil

dertuxmalwieder:
I don't use Git for almost all of my own tools, but I still use difftastic with my Fossil repositories. Here's how.


Wilfred:
Thanks :)

PR #517, opened 2023-05-17
Create draft release in pipeline

jankatins:
Github sends out an email on creating/publishing a release (Watch -> Custom -> releases) and that email contains whatever is at that point in the release description. As the pipeline currently creates a release without a description the email is mostly empty. As the changelog is added manually later, lets only create a draft release which can be "published" after adding the changelog. The release email then contains a proper changelog entry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,9 @@ jobs:
         env:
           # (required)
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Create a draft release so we can add the changelog before publishing it
+          # so the changelog is included in the release email
+          draft: true
 
   upload-assets:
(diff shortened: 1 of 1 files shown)


jankatins:
An alternative would be to parse out the changelog from the CHANGELOG.md file and add it during the release creating in the workflow, but I couldn't find how that file is maintained (manual?)

Wilfred:
I write CHANGELOG.md manually, and I've experimented with generating release notes (see `scripts/parse_changelog.sh`) but it needs work. Annoyingly GitHub markdown in repos ignores line wrapping, whereas release notes don't, so I tweak it by hand.

Wilfred:
Thanks, this is a definite improvement. I didn't realise people were getting emails on releases.

PR #522, opened 2023-05-26
Treat CUDA as C++

zasdfgbnm:


Wilfred:
Seems reasonable, thanks.

PR #557, opened 2023-08-17
Update semver version from 5.7.1 to 5.7.2

jereviikari:
Fixes CVE-2022-25883


jereviikari:
It seems there are plenty of other _semver_ library versions to be update in tree-setter lock files. That's a huge task.
This 5.7.1 one was the only one _trivy_ complained.

 _regex, minimist_ and _simple-get_ libraries should be updated also due to vulnerabilities.

Wilfred:
Thanks for the PR.

I'm only using the C/C++ code from the vendored parsers, so the remaining Cargo.lock files are completely unused.

You could consider updating the upstream repositories if you wish, and that'll get pulled in when I update the parsers. These files have no effect on difftastic though.

PR #559, opened 2023-08-19
fix(build): add rerun trigger for git HEAD and commit

eth3lbert:
- fix(build): add rerun trigger for git HEAD and commit

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -371,6 +371,16 @@ fn commit_info() {
         return;
     }
 
+    // Watch git HEAD and commit to trigger a rebuild
+    if PathBuf::from(".git/HEAD").exists() {
+        println!("cargo:rerun-if-changed=.git/HEAD");
+        let head_ref = std::fs::read_to_string(".git/HEAD").unwrap();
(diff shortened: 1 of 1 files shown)


Wilfred:
This seems correct, but it's going to re-run the whole build script on every commit. That seems like a big disadvantage when there are so many C libraries that are compiled by the build script.

eth3lbert:
Correct me if I'm wrong.
I got following output with `cargo build --timings`
<details>
<summary>Clean build graph</summary>
clean-build
</details>

<details>
<summary>Recompile</summary>
incremental
</details>

So I thought it still benefit from incremental compilation.

Wilfred:
I think your machine has a lot more CPUs than my thinkpad :)

Screenshot from 2023-09-18 08-38-00

You're correct that the other dependencies aren't rebuilt, but the build script time is by far the longest on my machine.

Wilfred:
Based on the additional build time, I'm going to close this. Thanks for taking the time to write the code and measure it :)

PR #562, opened 2023-08-26
Add options which allows us to use difft as Apache Subversion's external diff tool

futatuki:
svn,  Apache Subversion's command line client also has option to use external diff tool for `svn diff` subcommand.  However it assumes that diff tool has  diff compatible -L (label) and -u option.

This implements -L and -u options.


Wilfred:
Neat, it'd be nice to have svn support.

I'm OK with adding label options, but I don't like having multiple ways of setting the display mode.

Could we support this syntax similar to git support? If it's always the same arguments in the same order, we could allow those arguments without allowing `-u` in all circumstances.

futatuki:
Thank you for being interested in this PR.

`-u` option is passed only if the user does not specify any additional options.  This is done by [`svn_io_run_diff2`](https://subversion.apache.org/docs/api/latest/svn__io_8h.html#acabda4f4f591bec06ece8e35829a79f4)() C API function.   In the case,  `-u` options is always the first argument, 0 - 2 `-L LABEL` options is followed and then finally 2 paths,  however this order is not documented.

How about  hiding `-u` option from help and just ignore it in the argment processing ?

bmarwell:
I'd be interested!

bmarwell:
@futatuki can you rebase your diff, please?

futatuki:
Yes, I'm willing to do later, with pleasure.

bmarwell:
> Yes, I'm willing to do later, with pleasure.

Thanks! I am using your patch, it is really helpful. I hope @Wilfred sees this in time and can maybe integrate it :)

futatuki:
Rebase done.

futatuki:
This PR was broken again because of  backward incompatibility of clap. I'll try to catch up, later.

futatuki:
> This PR was broken again because of backward incompatibility of clap. I'll try to catch up, later.

done.

jcormier:
Interested in this being added.

zlatangao:
why this commit is not merged？ we need this for svn too

zlatangao:
@Wilfred can you help?

PR #574, opened 2023-09-21
Skip dirs specified in gitignore

### LINKED ISSUE #568: Add support for directory filtering
This is a really great tool!

I'm trying to use it outside of git, and apparently there's no easy way to filter out directories. There are some directories (`venv` for python, `node_modules` for js etc) which I definitely DO NOT want to diff. It's possible to implement this via reading a file (`.gitignore`?) or pass it in as an arg (or have it on by default). Are there any plans for this? I'm open to contributing if this sounds good and you need the extra hands 😄

nimrodkor:
Fixes #568 (hopefully)

Still learning my ways around Rust, so let it rip!


nimrodkor:
@Wilfred mind taking a look?

Wilfred:
Thanks for the PR! I agree this is a worthwhile feature, but globs and regex are different (e.g. * is a valid glob but not a valid regex).

Could you use the ignore crate to handle the .gitignore parsing here?

src/files.rs
```diff
@@ -234,23 +235,57 @@ pub fn guess_content(bytes: &[u8]) -> ProbableFileKind {
...
         .into_iter()
         .filter_map(Result::ok)
         .map(|entry| entry.into_path())
         .filter(|path| !path.is_dir())
+        .filter(|path| !is_ignored(ignored_dirs, path))
         .map(|path| path.strip_prefix(dir).unwrap().to_path_buf())
         .collect()
 }
 
+fn is_ignored(ignored_dirs: &Option<Vec<PathBuf>>, path: &PathBuf) -> bool {
```
Wilfred, line 249: FWIW it's generally better to take a slice argument in Rust unless you're expecting to push or pop items. In this case you could use `Option<&[PathBuf]>`.

Wilfred:
Looks good, thanks :)

PR #593, opened 2023-11-18
Sort by path before outputting directory diff so that the order is always the same

milandamen:
Previously, when doing a directory diff, the order in which the results were displayed was random, depending on thread execution and time it took for a file's diff.

This commit makes it so that the results are always shown in the same order.

It does this by diffing in parallel, then collecting the result, and then sorting it before printing it to the terminal.

This is the first time I have programmed in Rust. Any feedback is welcome.
I can also make a test for this, but as the order is random, the test might still pass even in the old situation.

I have run `cargo fmt` and `cargo test` and have tested it manually in a different codebase.


Wilfred:
This can make the UI way slower if there's any files with a lot of changes.

I'm not opposed to this feature, but could you put it behind a flag?

milandamen:
Thank you for the quick reply, and no problem. I will put it behind a feature flag. Do you want this functionality enabled or disabled by default?

Wilfred:
Disabled by default

milandamen:
@Wilfred I have made the changes you requested.

Wilfred:
Thanks :)

PR #598, opened 2023-11-29
Add --hidden and --no-ignore options to walk all files in directory

yuja:
The `ignore::Walk` excludes various files by default. It might makes sense if
user runs difft command manually, but difft can also be used as a directory
diff backend of version control tools. In that case, extra filtering is rather
problematic than useful.

The `--hidden` and `--no-ignore` flags are copied from ripgrep. We can add more
`--no-ignore-<kind>` as needed. For my use case, these two flags are good enough.
Alternatively, we can change the default to not exclude hidden (or dot) files.
That's sufficient for my needs, but I'm not sure if that's desired. The major
problem I have right now is that .gitignore changes are excluded (because it's
a "hidden" file.)

https://github.com/BurntSushi/ripgrep/blob/14.0.3/crates/core/flags/hiargs.rs#L889
https://github.com/BurntSushi/ripgrep/blob/14.0.3/crates/ignore/src/dir.rs#L586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Fixed an issue with line-based diffing where only the first line in a
 large changed region was highlighted. This was particularly noticeable
 when diffing brand new files.
 
+Added `--hidden` and `--no-ignore` options to not exclude ignored files from
+directory diffing.
+
 ### Display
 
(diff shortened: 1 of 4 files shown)


yuja:
Superseded by #654

PR #601, opened 2023-12-06
Use scoped thread API in directory diff frontend

yuja:
Just a code cleanup. As the minimum Rust version is now 1.63.0, we don't need
to manage the spawned thread manually. We can also remove the `AtomicBool`.


Wilfred:
Much nicer, thanks for the PR! :)

PR #618, opened 2024-01-08
Add tree-sitter-vhdl

arbrauns:
This adds support for [VHDL](https://wikipedia.org/wiki/VHDL) by way of https://github.com/JLeemaster/tree-sitter-vhdl.


Wilfred:
Thanks for the PR! Looks good to me.

The syntax highlighting is kinda weird: everything seems to be using `@error.foo` rather than highlighting e.g. keywords. Diffs still look sensible, just slightly less pretty :)

PR #634, opened 2024-02-01
Smali language support

evanrichter:
Following the guide for language support was pretty straight forward, thanks for the great docs!

This adds support for the Smali language, which is the human readable bytecode form of android APKs.

Before:

image

After:

image


amaanq:
Wow, it's cool to see people actually using the smali parser :grin:!

evanrichter:
> Wow, it's cool to see people actually using the smali parser :grin:! 

I added support to the Helix editor recently too! It's a wonder how much syntax highlighting helps

amaanq:
Smali is what got me into tree-sitter in the first place - I wanted syntax highlighting for it so bad in Neovim after moving away from vscode when reverse engineering android stuff :sweat_smile:

Wilfred:
Thanks! :)

PR #638, opened 2024-02-09
Add extra PHP file extensions to the language guesser.

RBotfield:
I've added in a few of the less commonly used PHP file extensions for the language guesser.

- `.phtml` is widely used on the Magento platform still to this day.
- `.php3`, `.php4`, `.php5`, and `.php7` are uncommon but can be found on some legacy systems.
- `.phps` is used for having PHP just return the source file, colourized, rather than processing it.


Wilfred:
These seem reasonable, but are you sure about .phtml? Based on the filename, I would expect a file to be primarily HTML rather than PHP.

RBotfield:
`.phtml` files are still PHP files parsed by the interpreter, just with HTML embedded in it. They're the views/templates/partials that are processed into HTML. An example using Magento, again: 
https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Catalog/view/frontend/templates/product/list.phtml
That product list template has ~114 out of its 165 lines written in PHP, the rest in HTML/JS, and for it to be at that kind of ratio or more is pretty common.

Wilfred:
Thanks for following up :)

PR #642, opened 2024-02-13
Support devicetree

blmaier:


Wilfred:
Merged, thank you!

There seems to be a weak preference for 'device tree' over 'devicetree' in my googling, so I've used that term.

PR #643, opened 2024-02-14
Allow JSON output to respect skip-unchanged

nimrodkor:
Today, JSON output doesn't respect the `skip-unchanged` param, and outputs all files regardless. This can result in unnecessarily big outputs. 

This fixes it


nimrodkor:
I don't think the failure is related to anything I changed @Wilfred ...

Wilfred:
Agreed, the failure was occurring on master too.

Thanks for the PR! :)

PR #654, opened 2024-03-01
fix: show diff for hidden but non-ignored files when walking directories

poliorcetics:
The first commit is a test showing that `difftastic` currently does not pick up hidden directories nor files.

The second fixes the issue.


gd4Ark:
I hope this question will be responded to as soon as possible, Thank you very much

Wilfred:
Perfect, thank you! :)

PR #663, opened 2024-03-21
Fix typo

jiripospisil:
Ripgrep is great too though.


Wilfred:
Oops! No prizes for guessing which other projects I looked at for examples of install instructions :).

Fixed in #665.

PR #673, opened 2024-03-22
from_source.md

arjunadeltoso:
Noticed from_source.md file is referenced in SUMMARY.md but was missing from the repo.

I just ran this sequence of commands to install difftastic on Ubuntu, hopefully they are general enough to cover most platforms.

### DIFF
--- a/manual/src/from_source.md
+++ b/manual/src/from_source.md
@@ -14,8 +14,10 @@ sudo apt install cargo
 
 ## Clone the repository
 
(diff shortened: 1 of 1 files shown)


Wilfred:
Argh, I moved the content and never checked in the new file! Thanks for the heads up, this should now be fixed.

By the way, if you'd like to contribute Ubuntu install instructions to the manual, that would be super helpful :)

PR #677, opened 2024-03-24
Fix Cargo Audit Errors by upgrading to clap 4 

Pi-Cla:
Clap 3 depends on atty which is an unmaintained crate with a [rustsec](https://rustsec.org/advisories/RUSTSEC-2021-0145) advisory. So we need to upgrade to clap 4(.2.1) which will raise the msrv to 1.67.1 at a minimum (later versions of clap 4 raise it even higher)

Also, just running `cargo update` bumped our transitive dependency on mio from 0.8.11 to 0.8.12 which fixed an actual serious [security vulnerability](https://rustsec.org/advisories/RUSTSEC-2024-0019)

---
[Migration guide to clap 4](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28)


Wilfred:
Thanks for the PR! I'm afraid I really don't want to upgrade the MSRV at this point. I'm actually hoping I'll be able to lower it back to 1.63 after the next release of tree-sitter, so I can continue to work towards Debian Stable support.

I'd happily accept a PR that bumps the mio dependency, and if you think the atty issue is a concern I'm happy to carry a patch for it too.

Wilfred:
Given that #678 is merged, I'm closing this. Thanks again for your PRs :)

PR #678, opened 2024-03-24
(Fix RUSTSEC2024-0019 by) run cargo update

Pi-Cla:
Just running `cargo update` on it's own bumped our transitive dependency on mio from 0.8.11 to 0.8.12 which fixed an actual serious [security vulnerability](https://rustsec.org/advisories/RUSTSEC-2024-0019)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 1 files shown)


Wilfred:
Benchmarking locally confirms that performance is unchanged, thanks for the PR!

Issue #689, opened 2024-03-29
Compilation Error on 3b049c2fd "cargo update"


egrieco:
After commit 3b049c2fd "cargo update" I'm seeing the following compilation error:

```
  cargo:warning=clang: warning: include path for libstdc++ headers not found; pass '-stdlib=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]
  cargo:warning=vendored_parsers/tree-sitter-lua-src/scanner.cc:2:10: fatal error: 'cwctype' file not found
  cargo:warning=#include <cwctype>
  cargo:warning=         ^~~~~~~~~
  cargo:warning=1 error generated.

  --- stderr


  error occurred: Command env -u IPHONEOS_DEPLOYMENT_TARGET "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-m64" "--target=x86_64-apple-darwin" "-mmacosx-version-min=10.7" "-std=c++14" "-I" "vendored_parsers/tree-sitter-lua-src" "-Wall" "-Wextra" "-o" "~/Repos/difftastic/target/debug/build/difftastic-a3b1e54c05e0af46/out/07332a6062849104-scanner.o" "-c" "vendored_parsers/tree-sitter-lua-src/scanner.cc" with args "c++" did not execute successfully (status code exit status: 1).
```

This the previous commit b3732b9eb compiles successfully.

Note: The error message above has been mildly edited. `/Users/<MY_USER>` was replaced with a tilde "~".

Wilfred:
Hmm, so the commit updated the `cc` crate, but I'm surprised it broke the build for you.

What platform are you on, with which compilers? This header is part of the stdlib apparently: https://en.cppreference.com/w/cpp/header/cwctype

FWIW mos ttree-sitter parsers are trying to move to be pure C, which might help too. The most actively maintained Lua parser seems to have no C++ at least: https://github.com/tjdevries/tree-sitter-lua

Wilfred:
I've updated the Lua parser and cc crate, and CI is still green. Feel free to reopen if you're still seeing issues.

PR #690, opened 2024-03-29
Redo of background colors to improve diff readability

egrieco:
This is a **complete rewrite** of #286 on top of the current version of `difftastic` (with the exception of the last cargo update which is causing a compilation error for me #689).

## Previous Issues

Hopefully this addresses the issues that @Wilfred mentioned in the last pull request:

* We probably don't need to make line numbers red/green with background highlighting: **no background highlighting behind line numbers**
* It looks like subword bold highlighting (the `6` in `Version 0.6`) and delimiters (the `(` before `reverse`) is missing: **not sure about this, please let me know if you still see this issue**
* I don't think you need cansi, you should be able to calculate lengths before applying colours: **got rid of cansi, mostly able to calculate proper lengths without the ANSI stripping step (except for inline view)**

## Themes

**This pull request lays the groundwork for theming** of `difftastic` in the future. The missing piece here is the ability to parse theme information from a config file. Internally, the various styles are stored in a `HashMap` and a lookup with fallback is performed when retrieving various values.

## Notes and Caveats

While this is a decent proof of concept there are probably a few issues in need of attention:

* Switched to [yansi](https://docs.rs/yansi/latest/yansi/index.html) for rendering colors in the terminal since it has this incredibly useful [wrapping](https://docs.rs/yansi/latest/yansi/index.html#wrapping) functionality. This makes "nesting" of ANSI colors feel possible.
* Inline view does not color lines all the way to the right side of the screen. **I'm having a rather difficult time tracking down where I need to add padding to make this happen.**
* Completely blank lines do not have any color. They only indication that they have changed is the color of their line number.
* Dark mode is not supported in themes. The idea is to use separate themes for dark and light mode.
* Header styles are not fully supported in the themes.
* `NO_COLOR` is not currently supported even though `yansi` supports [conditional styling](https://docs.rs/yansi/latest/yansi/struct.Condition.html#built-in-conditions) quite well.

## Reasoning

I'm creating this pull request now, in spite of it not being finished, since I'd like feedback on the above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,6 +292,7 @@ dependencies = [
  "unicode-width",
  "version_check",
  "wu-diff",
+ "yansi 1.0.1",
 ]
 
 [[package]]
@@ -739,7 +740,7 @@ dependencies = [
  "ctor",
  "diff",
  "output_vt100",
- "yansi",
+ "yansi 0.5.1",
 ]
 
 [[package]]
@@ -1237,6 +1238,12 @@ name = "yansi"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "09041cd90cf85f7f8b2df60c646f853b7f535ce68f85244eb6731cf89fa498ec"
+
+[[package]]
+name = "yansi"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfe53a6657fd280eaa890a3bc59152892ffa3e30101319d168b781ed6529b049"
 
 [[package]]
 name = "zerocopy"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -274,7 +274,6 @@ dependencies = [
  "line-numbers",
  "log",
  "mimalloc",
- "owo-colors",
  "predicates",
  "pretty_assertions",
  "pretty_env_logger",
@@ -661,12 +660,6 @@ checksum = "628223faebab4e3e40667ee0b2336d34a5b960ff60ea743ddfdbcf7770bcfb66"
 dependencies = [
  "winapi",
 ]
-
-[[package]]
-name = "owo-colors"
-version = "3.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1b04fb49957986fdce4d6ee7a65027d55d4b6d2265e5848bbb507b58ccfdb6f"
 
 [[package]]
 name = "parking_lot"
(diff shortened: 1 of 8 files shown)


gerhardol:
This looks better. 
Do you have a plan to create custom themes? To use git colors is one way I prefer.

egrieco:
> Do you have a plan to create custom themes? To use git colors is one way I prefer.

No. If you have more detail about what you'd like to see in custom theme support please share.

Right now, I have implemented this in a way that could support custom themes, specifically `yansi::Style`s are pulled from a lookup table implemented via a `HashMap`. The biggest missing pieces are

1. A format for storing themes on disk: Do we go with TOML, YAML, JSON, etc.
2. Parsing code to convert user specified theme info into actual `yansi::Style` types.

egrieco:
@gerhardol Is this what you have in mind?

* [Git - Git Configuration](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_colors_in_git)
* [Configuring git colors (Shallow Thoughts)](https://shallowsky.com/blog/programming/gitcolors.html)
* [Git Colors](https://jblevins.org/log/git-colors?ref=schaeflein.net)
* [Coloring your Git output](https://davidzych.com/coloring-your-git-output/)
* [Git Command Line color settings](https://www.schaeflein.net/git-command-line-color-settings/)

gerhardol:
More or less those links.
Git colors are defined in the doc: https://git-scm.com/docs/git-config#Documentation/git-config.txt-colordiffltslotgt
But you do not get the hardcoded defaults from Git what I  know.
Also, getting colors dynamically will be slow (some caching will have to be used).

But using the same names for the slot (with some additions and deletions) would simplify.
Maybe not this PR to define the available slots, but maybe the name of them.

Difftastic is more than Git, but as Git is dominant, at least configuration and documentation is simpler if the concepts ar the same.

In my case to use Difftastic as a difftool from within another tool, I would need some way to control tis at runtime, prefeably with environment variables. So a comma separated list with the Git names of the slots is my preference. A user may want other colors from the command line and use other theme colors when embedded in the app.

A sidetrack... 

Will try to build and try this branch myself

franroa:
Hi, what is the state of this PR? I think this is an amazing job and would be really having it merged

egrieco:
> Hi, what is the state of this PR? I think this is an amazing job and would be really having it merged

I keep getting blocked on one small issue or other that prevents this feature from being merged into difftastic.

For some reason I just can't seem to grok this particular codebase. Every time I've tried to implement this change properly and cleanly I'm always missing something. Just can't seem to find the "root" places in the code that need to be modified to handle all code paths cleanly.

It's been a while since I've taken a look. I'll try to take another crack at it soon.

franroa:
That sounds really great! This is the only thing keeping me from using difftastic. Thank you so much for your efforts!

PR #692, opened 2024-04-04
Much shorter version of dlog alias in git documentation

toy:
Use `-c` to set config instead of environment variable, so no need for building shell command


Wilfred:
Thanks :)

Issue #688, opened 2024-03-28
Panic index out of bounds in side_by_side.rs on go.sum


ofavre:
While running `git show --ext-diff` I noticed the following at the end of the diff:

```
[…]
go.sum --- 1/2 --- Text
  1 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
  2 cloud.google.com/go v0.110.0 h1:Zc8gqp3+a9/Eyph2KDmcGaPtbKRIoqq4YTlL4NMD0Ys=
  3 cloud.google.com/go v0.110.0/go.mod h1:SJnCLqQ0FCFGSZMUNUf84MV3Aia54kn7pi8st7tMzaY=
  4 cloud.google.com/go/compute v1.19.1 h1:am86mquDUgjGNWxiGn+5PGLbmgiWXlE/yNWpIpNvuXY=

go.sum --- 2/2 --- Text
.                                                                                                                     64 github.com/googleapis/enterprise-certificate-proxy v0.2.3/go.mod h1:AwSRAtLfXpU5Nm3pW+v7rGDHp09LsPtGY9MduiEsR9k=
.                                                                                                                     65 github.com/googleapis/gax-go/v2 v2.8.0 h1:UBtEZqx1bjXtOQ5BVTkuYghXrr3N4V123VKJK67vJZc=
.                                                                                                                     66 github.com/googleapis/gax-go/v2 v2.8.0/go.mod h1:4orTrqY6hXxxaUL4LHIPl6lGo8vAE38/qKbhSAKP6QI=
1 github.com/kylelemons/godebug v1.1.0/go.mod h1:9/0rRGxNHcop5bhtWyNeEfOS8JIWk580+fNqagV/RAw=                         67 github.com/kylelemons/godebug v1.1.0/go.mod h1:9/0rRGxNHcop5bhtWyNeEfOS8JIWk580+fNqagV/RAw=
2 github.com/naoina/go-stringutil v0.1.0 h1:rCUeRUHjBjGTSHl0VC00jUPLz8/F9dDzYI70Hzifhks=                              68 github.com/naoina/go-stringutil v0.1.0 h1:rCUeRUHjBjGTSHl0VC00jUPLz8/F9dDzYI70Hzifhks=
[...]

Issue #694, opened 2024-04-05
Error: "index out of bounds: the len is 20 but the index is 20"


Krinkle:
### Description

After several useful diffs, the output stops with this error. It's unclear to me whether this is a trailing error, or whether there would have been more output without this error. I worry the latter. 

```
<cut>

thread 'main' panicked at src/display/side_by_side.rs:488:58:
index out of bounds: the len is 281 but the index is 281
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at src/main.rs:284:38:
Receiver should be connected: SendError { .. }
```

### Test case

```shell
cd /tmp
git clone https://github.com/jquery/blog.jquery.com-theme blog && cd blog/ && git checkout -q 6710984735eb && cd ../
git clone --depth 1 -b v4.9.14 https://github.com/jquery/jquery-wp-content
difft blog/jquery jquery-wp-content/themes/jquery
```

### Version

macOS 13. Using Homebrew to install difft. Details at https://formulae.brew.sh/formula/difftastic and https://github.com/Homebrew/homebrew-core/blob/b5b224ebe92c94a69806ca1c3b90b9fbe7739e11/Formula/d/difftastic.rb.

```
$ difft --version
Difftastic 0.57.0 (built with rustc 1.77.1)
```

Krinkle:
```
RUST_BACKTRACE=1 RUST_BACKTRACE=full difft blog/jquery/ jquery-wp-content/themes/jquery/
```
```
thread 'main' panicked at src/display/side_by_side.rs:488:58:
index out of bounds: the len is 281 but the index is 281
stack backtrace:
   0:        0x102e85ae4 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h6e13e5c1f43a2491
   1:        0x102c9625c - core::fmt::write::ha2c1956860377338
   2:        0x102e8bf80 - std::io::Write::write_fmt::ha4b28c52ceab65a6
   3:        0x102e858d8 - std::sys_common::backtrace::print::h479f79ccc1585c96
   4:        0x102e9b09c - std::panicking::default_hook::{{closure}}::h3432bfbcced8e86b
   5:        0x102e9adb8 - std::panicking::default_hook::h0a224004d0f52354
   6:        0x102e9b4ec - std::panicking::rust_panic_with_hook::h0a46a8ecc3586611
[...]

Issue #681, opened 2024-03-26
Panic diffing text


davidselassie:
I'm not totally sure what about the first or second content is causing this, but I'm diffing two Markdown files (which are detected as text files) and get a panic. Using `Difftastic 0.56.1 (built with rustc 1.74.1)` on Ubuntu.

```
dsel@lemur ~/panictest (main) [101]> RUST_BACKTRACE=1 difft custom-connectors-old.md custom-connectors-new.md
custom-connectors.md --- Text
12 Also check out some of our built-in custom connectors  12 Also check out some of our built-in custom connectors 
.. in                                                     .. in
13 {py:obj}`bytewax.connectors` and looking at the source 13 {py:obj}`bytewax.connectors` and looking at the source
..  for how they                                          ..  for how they
14 are implemented.                                       14 are implemented.
thread 'main' panicked at /home/dsel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/difftastic-0.56.1/src/display/side_by_side.rs:505:34:
index out of bounds: the len is 14 but the index is 14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_bounds_check
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:180:5
   3: difft::display::side_by_side::print
   4: difft::print_diff_result
   5: difft::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I have tried to make a minimum reproduction. But removing seemingly any of the content from the new file does not cause a panic. Files here:

[custom-connectors-new.md](https://github.com/Wilfred/difftastic/files/14750848/custom-connectors-new.md)
[custom-connectors-old.md](https://github.com/Wilfred/difftastic/files/14750849/custom-connectors-old.md)

Wilfred:
Merging into https://github.com/Wilfred/difftastic/issues/688, thanks for the report!

PR #695, opened 2024-04-05
fix: out of bounds panic when diffing large number of words

### LINKED ISSUE #656: Crash on gauche commit
https://github.com/shirok/Gauche/commit/2eeca102b33a4ab7450d6fb5508642431fc46dc1

```
thread 'main' panicked at 'index out of bounds: the len is 49 but the index is 49', src/display/side_by_side.rs:505:25
fatal: external diff died, stopping at examples/spigot/README.adoc
```

### LINKED ISSUE #682: Panic diffing text file with --display inline (apparently interpreted as sql)
I hit a reproducible panic diffing some text file. Running on windows with difft.exe in the path, calling via WSL terminal.

```
difft.exe --version
Difftastic 0.56.1 (built with rustc 1.77.0)

wget https://raw.githubusercontent.com/FFmpeg/FFmpeg/master/COPYING.LGPLv2.1

RUST_BACTRACE=full difft.exe --display inline deleted_file ./COPYING.LGPLv2.1 commit_hash 10644 nul . .

thread 'main' panicked at C:\Users\Hannes.Franke\.cargo\registry\src\index.crates.io-6f17d22bba15001f\difftastic-0.56.1\src\display\inline.rs:129:38:
index out of bounds: the len is 0 but the index is 0
stack backtrace:
   0:     0x7ff6a90e2d71 - <unknown>
   1:     0x7ff6a8f3000d - <unknown>
   2:     0x7ff6a90df421 - <unknown>
   3:     0x7ff6a90e2b4a - <unknown>
   4:     0x7ff6a90e5379 - <unknown>
   5:     0x7ff6a90e4fe5 - <unknown>
   6:     0x7ff6a90e5a04 - <unknown>
   7:     0x7ff6a90e58e9 - <unknown>
   8:     0x7ff6a90e3469 - <unknown>
   9:     0x7ff6a90e55d6 - <unknown>
  10:     0x7ff6a98c6837 - tree_sitter_sql
  11:     0x7ff6a98c69e4 - tree_sitter_sql
  12:     0x7ff6a8f8726a - <unknown>
  13:     0x7ff6a8f72011 - <unknown>
  14:     0x7ff6a8f6c457 - <unknown>
  15:     0x7ff6a8f7a496 - <unknown>
  16:     0x7ff6a8fc7cfc - <unknown>
  17:     0x7ff6a90d86d5 - <unknown>
  18:     0x7ff6a8f724bc - <unknown>
  19:     0x7ff6a98c1fac - tree_sitter_sql
  20:     0x7ffd3e19257d - BaseThreadInitThunk
  21:     0x7ffd3f16aa58 - RtlUserThreadStart
```

JonathanxD:
*I dropped all the old text because it does not make sense anymore.*

The problem is basically that Rust's [`str::lines()`](https://doc.rust-lang.org/std/primitive.str.html#method.lines) will return the same number of lines regardless of the last character being a new line or not, and the assumptions made on *inline* and *side-by-side* display implementations depends on having a clear distinction of those cases.

~Those changes introduces a function that will chain an additional element when the string ends with a newline. I'm not aware of any function from stdlib that does what we need (and I couldn't find one by looking at the documentation), the closest one was `split`, but it has obvious implications (`\r\n`, for example).~

This does solves #688, #694, #682 and #681, I tested all the provided files and they're all working without any panics.

Update 2024-07-08:

I reworked on this patch and now it does not add a new line to the end of the string, therefore, all tests are passing, including the regression ones (which was failing before). In this new revision it fixes #682 again, which involves `--display=inline` and was not working anymore when I initially merged the changes from master into my branch.

I also tested another case that I found out and it fixes #656.

The new approach is more hacky: in one case it just removes the additional line range when the line is empty, and in another case it just checks if the index is within range before trying to access the index. However, unlike the previous solution, this one does not fail the regression tests.

### DIFF
--- a/src/display/side_by_side.rs
+++ b/src/display/side_by_side.rs
@@ -515,14 +515,16 @@ pub(crate) fn print(
                     None => vec![" ".repeat(source_dims.content_width)],
                 };
                 let rhs_line = match rhs_line_num {
-                    Some(rhs_line_num) => split_and_apply(
-                        rhs_lines[rhs_line_num.as_usize()],
-                        source_dims.content_width,
-                        display_options.tab_width,
-                        rhs_highlights.get(rhs_line_num).unwrap_or(&vec![]),
-                        Side::Right,
-                    ),
-                    None => vec!["".into()],
+                    Some(rhs_line_num) if rhs_line_num.as_usize() < rhs_lines.len() => {
+                        split_and_apply(
+                            rhs_lines[rhs_line_num.as_usize()],
+                            source_dims.content_width,
+                            display_options.tab_width,
+                            rhs_highlights.get(rhs_line_num).unwrap_or(&vec![]),
+                            Side::Right,
+                        )
+                    }
+                    _ => vec!["".into()],
                 };
 
                 for (i, (lhs_line, rhs_line)) in zip_pad_shorter(&lhs_line, &rhs_line)
--- a/src/line_parser.rs
+++ b/src/line_parser.rs
@@ -152,7 +152,22 @@ pub(crate) fn change_positions(lhs_src: &str, rhs_src: &str) -> Vec<MatchedPos> 
                 // have a very large number of words, don't diff
                 // individual words.
                 if lhs_words.len() > MAX_WORDS_IN_LINE || rhs_words.len() > MAX_WORDS_IN_LINE {
-                    for lhs_pos in lhs_lp.from_region(lhs_offset, lhs_offset + lhs_part.len()) {
+                    // If the last line is empty, `lhs_lines` will have one line less
+                    // than other sources, such as `LinePositions::from`.
+                    // This is a workaround for the off-by-one error caused by this disagreement
+                    // between `str::lines` and other sources used to calculate ranges.
+                    let lp_from_region = {
+                        let mut lp_from_region =
+                            lhs_lp.from_region(lhs_offset, lhs_offset + lhs_part.len());
+                        if let Some(last) = lp_from_region.last() {
+                            if last.start_col == 0 && last.end_col == 0 {
+                                lp_from_region.pop();
+                            }
+                        }
+                        lp_from_region
+                    };
+
+                    for lhs_pos in lp_from_region {
                         mps.push(MatchedPos {
                             kind: MatchKind::NovelWord {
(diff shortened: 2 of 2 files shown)


JonathanxD:
I ran a small script to compare the output of db281c682cb1e19f68392bf2bbffb27e48cba1be with my branch to find the cause of regression, it's all the trailing newlines that were not being printed before.

Wilfred:
Thanks for the pull request, and the clear explanation! This sounds like a regression from 8c004be87b5553809d1d865fc76adb754c2c3b55.

JonathanxD:
> Thanks for the pull request, and the clear explanation! This sounds like a regression from [8c004be](https://github.com/Wilfred/difftastic/commit/8c004be87b5553809d1d865fc76adb754c2c3b55).

I'm not sure if this is the cause of the regression because unless I'm missing something, the changes in this commit are just inlining the content of `split_on_newlines` on the call-site, which was already using `lines()`. I thought it could be different but after trying on [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=20bf7a0f07c20bada7eb1f2edce5c010), both versions yielded the same results.

@Wilfred Do you have any plans to merge my changes in the near future? If so, I can update the regression file so the pipeline succeeds. If you don't want to merge, I can use the examples provided in the reported issues to bisect which commit caused the regression and bring it to the discussion.

JonathanxD:
> What's the status of this PR? I'm running into this issue too.

Idk, everything is fine from my side. It depends on how @Wilfred wants to go from there.

I've been personally just building difftastic with my patches on top since without them difft would crash 90% of the time for my use-case.

JonathanxD:
I noticed that #731 does the same thing as I did (but slightly differently), which is to ignore the last empty line, and it will fix most of the out-of-bounds panics except for the [yaml](https://github.com/Wilfred/difftastic/issues/694#issuecomment-2066249883) one, which is caused by the same problem, but in a different location (mine "fixes" this one by skipping out of bounds indexes).

Wilfred:
Fixed: https://github.com/Wilfred/difftastic/issues/688#issuecomment-2241316356

Thanks for the PR and the helpful discussion!

PR #713, opened 2024-05-15
Fix off by one errors 

RGBCube:


Wilfred:
I don't think this is correct. Do you have a failing file example?

RGBCube:
> I don't think this is correct. Do you have a failing file example?

I used to, and it seems I forgot to follow up to this PR and show it 😅

Now it's long gone. I have experienced another panic in a while, so maybe it has already been fixed. I'll close this PR

RGBCube:
I now actually have a lot of test cases where difft panics: [difft-panic.tar.gz](https://github.com/user-attachments/files/16389536/difft-panic.tar.gz)

Each one of these files panic with `Difftastic 0.58.0 (built with rustc 1.78.0)`, assuming you diff the correct files (.a and .b with the same stem)

@Wilfred this might be of some help

Issue #704, opened 2024-04-23
Inline diff is ignoring unchanged lines between hunks


olivergondza:
## A description of the issue. A screenshot is often helpful too.

Inline diff (side-by-side works well) skips unchanged lines producing confusing diffs:

```$ difft --display inline --context 5 before.txt after.txt```

![Screenshot from 2024-04-23 13-08-50](https://github.com/Wilfred/difftastic/assets/206841/830c577c-e910-4d3b-bffe-61bccbb2bc5d)

side-by-side is showing the `try` and `finally` lines ok:

```$ difft --context 5 before.txt after.txt```

![Screenshot from 2024-04-23 13-09-04](https://github.com/Wilfred/difftastic/assets/206841/d3b2a6a0-71b3-4bc0-a73a-aed8205b9818)

## A copy of what you're diffing. If you're diffing files, include the before and after files. If you're using difftastic with a VCS repository (e.g. git), include the URL and commit hash.

Using [before.txt](https://github.com/Wilfred/difftastic/files/15075795/before.txt) and [after.txt](https://github.com/Wilfred/difftastic/files/15075796/after.txt).

Note those are groovy files (that difftastic handles as text). But even renamed to `.java`, the inline diff have same problem.

## The version of difftastic you're using (see `difft --version`) and your operating system.

Difftastic 0.58.0 (f52ca706f 2024-04-09, built with rustc 1.65.0)
Linux arch 6.8.4-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 05 Apr 2024 00:14:23 +0000 x86_64 GNU/Linux

mybuddymichael:
Agreed that `--display inline` is misleading right now. It makes the diff look like something it's not. I'd recommend people not use it until this is fixed.

Jaeiya:
Any updates on this issue? I'd love to use this with lazygit, but because inline display is kind of required, I've been holding off until this bug is fixed.

PR #730, opened 2024-06-13
Add insta-based snapshot tests that cover side-by-side format

yuja:
I'm going to fix inline diff issue #704, and found that the inline format isn't
covered by CLI tests. I could duplicate compare_all.sh to test various formats,
but it's uneasy to review that a new formatting is better than the previous
version. I think snapshot testing will help guarantee the output quality without
increasing maintenance burden too much.

If we like the idea, maybe we can migrate some of compare_all.sh tests to the
insta-based ones.

Note that this will add ~10sec to "cargo test" runs.


Wilfred:
Thanks for the PR! It's a neat idea, but I don't think it's a good fit for difftastic.

Diffs of diff output will quickly get confusing, so I prefer just knowing if output has changed and then eyeballing the changes. It also helps catch issues with coloured regions changing, which aren't really captured by regression tests that show the text output.

Issue #620, opened 2024-01-10
Git diff paths are confusing when using `git difftool` (but not `git diff`)


poliorcetics:
## Issue

```bash
cd /tmp/
cargo new --bin repro
cd repro
git add .
git commit -m "First commit"
echo 'fn main() { println!("Repro for difftastic"); dbg!(42); }' > src/main.rs
git difftool --tool difftastic  # See https://difftastic.wilfred.me.uk/git.html
```

This shows:

![Screenshot 2024-01-10 at 09 24 49](https://github.com/Wilfred/difftastic/assets/7951708/064963fb-f2db-4c9a-bbc9-26feea175aea)

Note the path: `main.rs` instead of `src/main.rs`. Using `GIT_EXTERNAL_DIFF=difft git diff` does not have this problem for some reason ?

## Solution

I replaced the `[difftool "difftastic"].cmd` with `echo "$LOCAL" "$REMOTE"` and it shows this: `/tmp/nix-shell.Pyzfqe/git-blob-OVxVxg/main.rs src/main.rs`

`/tmp/nix-shell.Pyzfqe` is `$TMPDIR` for me so really git is passing `LOCAL="$TMPDIR/git-blob-OVxVxg/main.rs"`.

So here is what I propose:

1. When two paths are given
2. And they're both files
3. And one of them starts with `$TMPDIR` and the other doesn't (when $TMPDIR is unset, `/tmp/` is used)
4. And the directory just under said TMPDIR starts with `git-blob-`
5. Then the path used in `difftastic` output should be the other path

I think this will work on both Linux and macOS. I don't know how git windows works

Note: the second path given to `difftool` is always from the root of the repo, we could make it relative to the current dir (when possible, falling back to the original path) to make it more intuitive, like a simple `git diff` works)

## Meta

- Difftastic: 0.54.0 (built with rustc 1.75.0)
- OS: macOS 14.1 (23B74)

poliorcetics:
Workaround: using `cmd = difft --tab-width 4 "$BASE" "$LOCAL" "hash" "mode" "$REMOTE" "hash" "mode"` works well enough, though it is an hack based on how difftastic parses args and how "git diff" passes arguments

Wilfred:
Thanks, I agree this is worth fixing and your heuristic sounds good.

PR #734, opened 2024-07-12
feat: use $MERGED if present in environment, as the display path

gthb:
In #620 the display path was fixed to be based on the right-hand-side filename in the case that the left-hand-side filename is a git temporary file.

But that does not help for the case of git invocations comparing two git commits, such as:

* `git dft main...`
* `git dft branch1...branch2`

... because, in those cases, both the lhs and the rhs path are git temporary files, obscuring the real path to the file within the git repository.

The same applies if a file is removed or added, in which case one of the paths is a git temporary file and the other is `/dev/null`.

In those cases, git makes the real path available as the `MERGED` shell variable, for the purposes of constructing a difftool command line in git configuration.

So in Difftastic, let the real path be specified by the `MERGED` environment variable (overriding the lhs and rhs paths), and suggest `export`-ing that into Difftastic's environment in the git difftool configuration.


Wilfred:
Maybe difftastic should just respect MERGED when invoked from git, rather than expecting a REAL_PATH environment variable? Happy to merge this if you make that change :)

gthb:
> Maybe difftastic should just respect MERGED when invoked from git, rather than expecting a REAL_PATH environment variable? Happy to merge this if you make that change :)

That's what I tried first, hoping to not need any `.gitconfig` change ... but sadly the `MERGED` that git provides is just a _shell_ variable, in the shell that invokes the difftool. So the difftool command line in `.gitconfig` must explicitly make it an environment variable.

I renamed it to `MERGED` just now, in https://github.com/Wilfred/difftastic/pull/734/commits/c087c661c58140d56eccb1ed82e09d3bc828407e, but it still needs to be there in `.gitconfig`:

```diff
 [difftool "difftastic"]
-        cmd = REAL_PATH="$MERGED" difft "$LOCAL" "$REMOTE"
+        cmd = MERGED="$MERGED" difft "$LOCAL" "$REMOTE"
```

That's why I chose the different name `REAL_PATH`, to avoid user confusion (“but I can just skip this, right?”).

gthb:
> That's why I chose the different name `REAL_PATH`, to avoid user confusion (“but I can just skip this, right?”).

Just realized that this would be less confusing if the `.gitconfig` example explicitly uses `export MERGED &&` so did that in https://github.com/Wilfred/difftastic/pull/734/commits/46e835b88c05ec4b035a49fdca087c0a652c182c

gthb:
Not sure what causes these CI test failures. I can't reproduce them locally, even using the 1.65.0 toolchain that CI uses:
```sh
rustup toolchain install 1.65.0 --target aarch64-apple-darwin --profile minimal --no-self-update
cargo +1.65.0 test --target aarch64-apple-darwin 
```

I do get several dozen copies of this in the terminal when building (before and after switching toolchain):
```
warning: xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation
warning: xcrun: error: unable to lookup item 'PlatformVersion' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
```
... which I've been ignoring. But I was getting that also when I created the pull request, and the CI tests did pass then, so not sure that's relevant.

gthb:
I noticed that this was missing the case where a file is removed or added, in the commit span being compared, so lhs or rhs is `/dev/null`. Moved this case out one level to cover that, in https://github.com/Wilfred/difftastic/pull/734/commits/53b899e31e20eca016a3d820fbc1b4763fc3f05c

gthb:
Could you consider merging this (or telling me what I need to do with it first :) ) so that I don't have to maintain a local build for this fix?

Wilfred:
Crikey, difftool is generally not what we want (git doesn't pass rename information), but external diff tools crash unless you're on a recent git version.

I've overhauled the git configuration page in the manual, and I think it fixes the issues you've encountered. Let me know if not :)

gthb:
Looks like it does!

PR #735, opened 2024-07-17
Add shell completions using clap_complete

yedayak:
This uses the last 3.x clap_complete version, which matches the currently used clap version (regarding MSRV).

Note the TODO I left, not sure what to do about that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,6 +168,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Wilfred:
Thanks for taking the time to create a PR, and sorry for the slow response!

Whilst the vast majority of difft usage is (I presume) `difft old.js new.js` or using it in git, there are a bunch of flags when you do `--help`. So this is probably a small but real improvement.

If you'd like to rebase this I'd be happy to merge it :)

yedayak:
I rebased but there is some issue with the package version and rust edition, didn't look into it. In general I tested this again and it seemed to have an unfortunate side effect of not completing paths by default, only if you start a path. For example type `difft <TAB>` only completes flags, you have to do `difft ./<TAB>` to see files. This is testing in bash, might be better in fish/zsh.
I think this arguably might be a worse experience, open to input.

PR #736, opened 2024-07-18
Update the cc package to 1.1.5

yedayak:
This includes the fix to https://github.com/rust-lang/cc-rs/issues/909 which spams warnings on ubuntu 22.04


Wilfred:
Thanks 😊

PR #754, opened 2024-08-20
Include manpage in published crate

michel-slm:
```
$ cargo package --list --allow-dirty | grep difft
difft.1
```


michel-slm:
Actually, on second thought this will cause the manpage to be copied into the shared registry during Fedora packaging, which is probably not what we want

PR #756, opened 2024-08-25
feat: add Tact language

novusnota:
More about the language: https://tact-lang.org

- Added the vendored [tree-sitter-tact](https://github.com/tact-lang/tree-sitter-tact)
- Configured the build
- Updated `tree_sitter_parser.rs`
- Updated `guess_language.rs`
- Added syntax highlighting queries
- Added a regression test

Followed the [manual](https://difftastic.wilfred.me.uk/adding_a_parser.html), hope everything's correct here :)


novusnota:
@Wilfred Thanks for the response and your time! Sure, pluggable parsers definitely seem like a better option than to vendor them in via git.

Perhaps, difftastic could take the approach of NeoVim and Helix: one only needs to specify the basic info about the parser, like does it have `scanner.c` or only `parser.c` (default), the target repo with it, and the SHA of the target commit. Then, they can fetch the parsers and build them on demand with any C compiler on the system, producing a bunch of shared `.so` objects. In the end, they only bundle the queries (a small bunch of small text files) and allow users to override the queries locally too, if needed.

That approach has its downsides of course, but it works well enough in practice :)

PR #759, opened 2024-08-31
Parse pdm.lock and uv.lock as TOML

idea-list:
PDM and uv are package managers for Python and both use TOML for lock files ([linguist](https://github.com/github-linguist/linguist/blob/ebbedf06dc18e929104a4a3573ac43b43298a761/lib/linguist/languages.yml#L7164-L7166)).


Wilfred:
Thanks :)

PR #760, opened 2024-09-01
[docs] Remove unnecessary double quotes in Git alias

injust:
This makes the config option identical to running `git config --global alias.dlog '-c diff.external=difft log -p --ext-diff'`.


Wilfred:
Thanks 🙂

PR #780, opened 2024-10-29
Avoid random ordering of lines in JSON output.

UhhhWaitWhat:
Pretty much what it says on the tin.

As the lines are stored in a HashMap, calling `into_values()` gives us a random order of lines.
My solution here is definitely not pretty but was the simplest thing that worked without putting too much thought into the code. I hope thats fine. And thank you for the tool :)


Wilfred:
I went for a different solution, but thanks for the report and PR :)

PR #785, opened 2024-11-18
Improve diffs inside Elixir sigils

### LINKED ISSUE #784: Changes in Elixir HEEx sigils are not highlighted
When changing code inside an embedded HEEx sigil on Elixir, difftastic is not highlighting the changes but the whole sigil.

I have the [following code](https://github.com/elixir-error-tracker/error-tracker/blob/428f4c78d7450086f01958465f89bd511edfadbc/live.exs#L40-L58):

```elixir
  def render(assigns) do
    if assigns.count == 5 do
      raise "Crash on render"
    end

    ~H"""
    <span><%= @count %></span>
    <button phx-click="inc">+</button>
    <button phx-click="dec">-</button>
    <button phx-click="error">Crash on handle_event</button>

    <.link href="/?crash=mount">Crash on mount</.link>
    <.link patch="/?crash=handle_params">Crash on handle_params</.link>

    <style type="text/css">
      body { padding: 1em; }
    </style>
    """
  end
```

And I change it to (just added the `new` class to the first `<span>` element):

```elixir
  def render(assigns) do
    if assigns.count == 5 do
      raise "Crash on render"
    end

    ~H"""
    <span class="new"><%= @count %></span>
    <button phx-click="inc">+</button>
    <button phx-click="dec">-</button>
    <button phx-click="error">Crash on handle_event</button>

    <.link href="/?crash=mount">Crash on mount</.link>
    <.link patch="/?crash=handle_params">Crash on handle_params</.link>

    <style type="text/css">
      body { padding: 1em; }
    </style>
    """
  end
```

When using basic git diff it highlights the changed line. On the other hand, difftastic highlights the whole block:

image

It is important to note that diffs are properly highlighted on HEEx files. The problem appears only on embedded HEEx blocks.

I am using latest difftastic on MacOS:

```
Difftastic 0.61.0

Toolchain: 1.82.0
System:    macos aarch64
```

crbelaus:
This pull request makes the following changes:
1. Updates the vendored [tree-sitter-elixir](elixir-lang/tree-sitter-elixir) package
2. Adds a few Elixir example files so we can compare diffs before and after the changes
3. Updates the Elixir parser to consider sigils as atom nodes

This is the result (before and after):

image

Notice how after the changes, the modifications inside the `~H` sigil are highlighted when previously they were not. This fixes #784


crbelaus:
Looks like updating the [tree-sitter-elixir](elixir-lang/tree-sitter-elixir) subtree has included many commits from that repository. Let me know if you prefer me to squash them into a single commit instead.

Wilfred:
Merged in b0a4ddc7ed58b76b63119b2b983c7a3abd1a4cd6, thanks for the fix and the regression test :)

PR #789, opened 2024-11-25
Stop using deprecated macos-12 runner

crbelaus:
The `macos-12` runner is [deprecated](https://github.com/actions/runner-images/issues/10721) and causing pipelines to fail.

This PR updates it to `macos-14-large` which is still [Intel-based](https://github.com/actions/runner-images).


crbelaus:
Oh wait, looks like `macos-14-large` is not available in the free tier 🤦‍♂️

PR #813, opened 2025-02-10
Avoid race condition

bmwiedemann:
There seems to be some race-condition in `build.rs`
that causes random build results to be produced for `/usr/bin/difft`.

This change seems to have negligible influence on overall build time.

If you make a better patch, I can test it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -249,7 +249,7 @@ fn main() {
         println!("cargo:rerun-if-changed={}", parser.src_dir);
     }
 
-    parsers.par_iter().for_each(|p| p.build());
+    parsers.iter().for_each(|p| p.build());
     commit_info();
 
     if let Some((version, _, _)) = rustc::triple() {


Wilfred (changes requested):
Thanks for the PR!

Can you share more information? Are you getting invalid binaries, or valid but not bit-for-bit identical binaries? What platform are you building on?

I'd also be surprised if building all the C libraries in parallel wasn't still a big improvement for build time. Try removing `.flag_if_supported("-Wno-unused-parameter")` from build.rs and see if the build time is comparable with and without your change.

bmwiedemann:
It produced different binary results. Probably all working, but did not test. I think, some elements were ordered differently.

You can use https://github.com/bmwiedemann/reproducibleopensuse/blob/master/filterdiff with 'strings' and difft binaries as params. (and the diff is probably more readable with `lto = "off"` )

I measured build times on a 2x16 core 2.6GHz host with build VMs:
without patch on 1-core-VM: 657s
with patch on 1-core-VM: 654s
without patch on 4-core-VM: 142s
with patch on 4-core-VM: 175s

So the patch does slightly speed up single-core builds but slow down multi-core builds. If we can avoid the variations in other ways, that would be good.

`-Wno-unused-parameter` seems to not make a difference.

Wilfred:
Could you try setting `RAYON_NUM_THREADS` to 1? That might be sufficient to fix this.

bmwiedemann:
Indeed, this seems to produce the same binaries as with this PR applied.

PR #816, opened 2025-02-13
Update installation.md for cargo

antoine2711:
Add the cargo installation and URL.

Regards, Antoine


Wilfred:
Thanks for the PR! I think this is already documented here: https://difftastic.wilfred.me.uk/from_source.html

PR #842, opened 2025-06-05
add verilog, systemverilog support

bryceberger:


Wilfred:
Overall this looks great! Could you update the changelog for this new language?

Also, are you sure there are no nodes that should be atoms? Try dumping the tree sitter AST for a program with complex string and integer literals to check.

bryceberger:
I wasn't quite sure what the atoms were for. Correct me if I'm wrong ---

If a sample parsed like the following:

```
0x123
---
(literal [0, 0] - [0, 4]
  (base [0, 0] - [0, 1])
  (value [0, 2] - [0, 4]))
```

That is, having a separate node for the "base" and the "value".

Then, having `literal` be in the list of atoms would have difft treat `0x123` as entirely different from `0o123`, even though only the `base` node changed.

Assuming that understanding is correct, I don't think any nodes need to be explicitly marked as atoms. Complex literals like `12'habc_123` have the following parse tree:

```
12'habc_123
---
(expression [3, 11] - [3, 21]
  (primary [3, 11] - [3, 21]
    (primary_literal [3, 11] - [3, 21]
      (integral_number [3, 11] - [3, 21]
        (hex_number [3, 11] - [3, 21])))))
```

Multiple nodes, but they all share the entire span.

Time literals are separated, but I think that makes sense --- changing 10ns to 10ps will show `ns` changed to `ps`, but the number stayed the same.


```
10ns
---
(time_literal [0, 11] - [0, 14]
  (unsigned_number [0, 11] - [0, 12])
  (time_unit [0, 12] - [0, 14]))
```

Wilfred:
Merged, thanks!

I've tried to clarify the atom nodes rationale in 6b522b867c082cae5c027d7c68a5d25baed82b65, please let me know if that helps clarify things.

Based on your examples above, the only nested literal is `integral_number`, so I've added that to the Verilog config.

PR #843, opened 2025-06-05
compare_all.sh: get difft binary using `cargo metadata`

bryceberger:
I have cargo configured to have a global target directory (in my case, at /tmp/cargo). The output directory for cargo builds is not guarunteed to be at ./target.

This could alternatively be solved by the user with strange defaults (me) making a symlink like `ln -s /tmp/cargo ./target`, which doesn't require having `jq` installed.

### DIFF
--- a/sample_files/compare_all.sh
+++ b/sample_files/compare_all.sh
@@ -16,6 +16,7 @@ set -e
 
 echo "==> Building difftastic"
 cargo build --release
+difft_bin=$(cargo metadata --format-version 1 | jq -r '.target_directory')/release/difft
 
 # Set language so we expand globs in a consistent order regardless of
 # locale (e.g. on GitHub actions).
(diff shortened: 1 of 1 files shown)


bryceberger:
Having the user (me) add a symlink, and not add a test-time dependency on `jq`, is a perfectly fine workaround in my mind. Feel free to close with that as the recommended solution.

Wilfred:
This seems reasonable to me.

(Incidentally, why did you change your output directory?)

bryceberger:
I changed my `cargo` target directory to `/tmp/cargo` so that I didn't have a bunch of different `target/` directories sitting around. I can deal with the total disk usage, but having it scattered in 10+ different places is annoying.

I think there are other tools to deal with this, but this seemed like a simple enough solution to me. It's not all great --- building two things at once, each cargo process wants to lock the same build directory. I think building with different cargo versions can clobber some artifacts. Overall, manageable for me since I mostly work on one project at a time with one cargo version.

Separately, I've found that having a separate rust-analyzer target directory is also helpful --- mine is at `/tmp/rust-analyzer`. Stops r-a analyzing from blocking cargo builds.

PR #844, opened 2025-06-06
Update gleam parser

llakala:
Can confirm it builds locally, and now functions without parse errors.


llakala:
Fixed the sample - they were using the old `assert Ok(_) =` syntax, which is no longer valid Gleam. The tests definitely aren't a great example of WORKING Gleam - most of the imported libraries no longer exist - but at least now they're parsable.

llakala:
The test now only "fails" because of the syntax change.  I ran a diff locally of the `difft` output before and after this PR, and this was the result:
image

Wilfred:
Could you update the regression test output? I agree that change looks totally reasonable -- sometimes parser changes slightly change where colours start and stop, which changes the hash of the output.

llakala:
Sorry, could you clarify exactly how to do that?

Wilfred:
Assuming you have https://github.com/casey/just installed, do the following:

```
$ just compare
...
==> Result
Output changed. If this was expected, update with:
  cp sample_files/compare.result sample_files/compare.expected
```
```

llakala:
Gotcha, will do tomorrow!

llakala:
Done!

Wilfred:
Thank you :)

PR #849, opened 2025-06-17
Switch to tree-sitter-rust-orchard, fixing attribute attachment in Rust

### LINKED ISSUE #35: Rust attribute changes are not ideally associated
Using the [indexmap](https://github.com/bluss/indexmap) repo, I tried the following command:

    GIT_EXTERNAL_DIFF=difftastic git show --ext-diff 49d430808ced6d1824ec15d621acd6b0e8367c19

and got this output near the end:

image

Notice how the `#[test]` attribute is interpreted, where the following one is considered to be an insertion, instead of the one that's on the new function. This kind of thing is very common with line-based diffs, although plain `git show` actually does get this right!

image

wetneb:
This proposes to update the Rust grammar, to benefit from the following improvements:
- fix the attachment of attributes (and doc comments) to functions (and many other types of nodes).
- fixes the parsing of various stable language constructs
- add support for newer language constructs (including unstable ones).

See [the changelog](https://codeberg.org/grammar-orchard/tree-sitter-rust-orchard/releases) for details about those improvements.

Those are improvements I made for mergiraf but I figured they might also be useful for difftastic, and indeed I found that this change indeed fixes #35.

Unfortunately the tree-sitter-rust maintainers rarely have capacity to review external contributions, hence it being a fork called [`tree-sitter-rust-orchard`](https://codeberg.org/grammar-orchard/tree-sitter-rust-orchard), which is [available on crates.io](https://crates.io/crates/tree-sitter-rust-orchard). The lack of capacity to review seems to be a recurrent problem according to the PR backlog, not just a passing one, but should it change, I would of course submit my changes upstream to have them integrated there too. In the meantime, the fork welcomes contributions and co-maintainers.


Wilfred:
Thanks! This largely looks good.

I noticed that block_outer_doc_comment isn't handled as a comment in this PR, but you've just fixed it: https://codeberg.org/grammar-orchard/tree-sitter-rust-orchard/commit/d52d4268f15b9c715cd52010cc909f92511610d3 :)

I did also notice that there's an unnecessary token, so I've filed https://codeberg.org/grammar-orchard/tree-sitter-rust-orchard/issues/70. I think it's harmless for difftastic, which flattens AST subtrees which are comments anyway.

Wilfred:
I see the tests are all passing with 0.13, so merging. Thanks for the PR! :)

PR #856, opened 2025-07-04
Add support for D 

baruch:


baruch:
Resolved conflicts

Wilfred (changes requested):
Thanks for the PR! I'd be up for adding D support, but I'm moving to using crates.io libraries for all the parsers. Could you switch to using https://crates.io/crates/tree-sitter-d instead? It's the same library.

PR #868, opened 2025-08-02
Fix GitHub logo reflow

Shraymonks:
- Set a fixed width on the GitHub logo to prevent reflow while loading
- Set `pointer-events: none;` so clicking the logo still links correctly

https://github.com/user-attachments/assets/68e942d9-4a5b-4cb1-b063-954b25329787

https://github.com/user-attachments/assets/90f88fc7-3ea5-4e19-8066-5f3641e6037f

### DIFF
--- a/homepage/home_static/style.css
+++ b/homepage/home_static/style.css
@@ -69,7 +69,8 @@ a {
 }
 
 .github-logo {
-  padding-right: 5px;
+  margin-right: 5px;
+  pointer-events: none;
 }
 
 .card-set .card {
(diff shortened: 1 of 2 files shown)


Wilfred:
Thanks, this seems reasonable. I've not noticed this myself though, is it browser specific?

Shraymonks:
It’s possible that this might not be noticeable depending on the browser or network, but the [spec](https://www.w3.org/TR/css-sizing-3/#intrinsic-sizes) does define 300x150 as the default dimensions.

PR #874, opened 2025-08-29
Add support for gitattributes

### LINKED ISSUE #466: .gitattributes definitions ignored by difft used with git diff
After configure git to use difftastic as diff tool
```
#~/.gitconfig
[diff]
	external = difft
```
And set an specific file as binary, it is treated as text by git diff.
```
# .gitattributes
.yarn/releases/yarn-3.3.1.cjs binary
```

the output of git diff indicates

`.yarn/releases/yarn-3.3.1.cjs --- Text (exceeded DFT_BYTE_LIMIT)`

and includes the file in the output of the diff

WGH-:
Runs `git check-attr` to get the files git attributes, and treats the file as binary if the attribute value is `"unset"`, matching the built-in git diff.

Fixes #466

Idea-by: @anuramat (https://github.com/Wilfred/difftastic/issues/466#issuecomment-3229267882)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,6 +47,7 @@ mod diff;
 mod display;
 mod exit_codes;
 mod files;
+mod gitattributes;
 mod hash;
 mod line_parser;
 mod lines;
@@ -75,6 +76,7 @@ use crate::files::{
     guess_content, read_file_or_die, read_files_or_die, read_or_die, relative_paths_in_either,
     ProbableFileKind,
 };
+use crate::gitattributes::{check_diff_attr, DiffAttribute};
 use crate::parse::guess_language::{
     guess, language_globs, language_name, Language, LanguageOverride,
 };
@@ -414,8 +416,11 @@ fn diff_file(
     let (mut lhs_src, mut rhs_src) = match (
         guess_content(&lhs_bytes, lhs_path, binary_overrides),
         guess_content(&rhs_bytes, rhs_path, binary_overrides),
+        check_diff_attr(Path::new(display_path)),
     ) {
-        (ProbableFileKind::Binary, _) | (_, ProbableFileKind::Binary) => {
+        (ProbableFileKind::Binary, _, _)
+        | (_, ProbableFileKind::Binary, _)
+        | (_, _, Some(DiffAttribute::AssumeBinary)) => {
             let has_byte_changes = if lhs_bytes == rhs_bytes {
                 None
             } else {
@@ -434,7 +439,7 @@ fn diff_file(
                 has_syntactic_changes: false,
             };
         }
-        (ProbableFileKind::Text(lhs_src), ProbableFileKind::Text(rhs_src)) => (lhs_src, rhs_src),
+        (ProbableFileKind::Text(lhs_src), ProbableFileKind::Text(rhs_src), _) => (lhs_src, rhs_src),
     };
 
     if diff_options.strip_cr {
--- a/src/gitattributes.rs
+++ b/src/gitattributes.rs
@@ -1,0 +1,166 @@
+use std::path::Path;
+use std::process::Command;
+
+/// The result of checking file's `diff` and `binary` attributes with `git check-attr`.
+/// See man gitattributes, specifically **Generating diff text** section.
+#[derive(Debug, PartialEq)]
+pub(crate) enum DiffAttribute {
+    /// The file type (text vs binary) will be autodetected.
+    ///
+    /// This is the default setting, when `diff` attribute is not present.
+    Unspecified,
+
+    /// The file should be treated as text.
+    ///
+    /// Corresponds to set `diff` attribute, unless `binary` attribute is set.
+    AssumeText,
+
+    /// The file should be treated as binary, and no diff will be shown.
+    /// Sometimes this option is set for generated text files.
+    ///
+    /// Corresponds to explicitly unset `diff` attribute (`-diff`), or when `binary` attribute is
+    /// set.
+    AssumeBinary,
+}
+
+/// Runs `git check-attr diff binary` to get the diff and binary attributes of the path. Returns
+/// [`Option::None`] when either `git` is not available, file is not inside git directory, or
+/// something else went wrong.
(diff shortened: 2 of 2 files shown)


WGH-:
This is somewhat quick-and-dirty solution. Perhaps, we we shouldn't run `git check-attr` so eagerly. In any case, I want to hear some opinions.

src/git.rs
```diff
@@ -0,0 +1,63 @@
+use std::path::Path;
+use std::process::Command;
+
+/// Corresponds to the diff attribute. See man gitattribute.
```
Wilfred: I think you mean `man gitattributes`? There's no singular version that I can see.

src/git.rs
```diff
@@ -0,0 +1,63 @@
...
+pub(crate) enum DiffAttribute {
+    Set,
```
Wilfred: Would you mind adding doc comments to these? Even a brief phrase copied from the man page would be a big help. I think `Other` is something you've added, and isn't defined in git?

src/git.rs
```diff
@@ -0,0 +1,63 @@
...
+    Other,
+}
+
+impl From<&str> for DiffAttribute {
+    fn from(s: &str) -> Self {
+        match s {
+            "set" => Self::Set,
+            "unset" => Self::Unset,
+            "unspecified" => Self::Unspecified,
+            _ => Self::Other,
```
Wilfred: I guess the other option would be to use TryFrom so you don't need this catch-all Other value.

src/git.rs
```diff
@@ -0,0 +1,63 @@
...
+                Some(value) => match std::str::from_utf8(value) {
+                    Ok(s) => return Some(s.into()),
+                    Err(err) => {
+                        warn!("invalid diff attribute value: {err}");
+                    }
+                },
+            }
+        }
+        Err(err) => {
+            warn!("failed to execute git: {err}");
```
Wilfred: These should be debug logging I think, especially this one. You can run `difft` in a directory that isn't a git repository, or even on a machine that doesn't have git installed.

src/git.rs
```diff
@@ -0,0 +1,63 @@
...
+            "unspecified" => Self::Unspecified,
+            _ => Self::Other,
+        }
+    }
+}
+
+/// Runs `git check-attr diff` to get the diff attribute of the path. Returns
+/// [`Option::None`] when either `git` is not available, file is not inside git
+/// directory, or something else went wrong.
+pub(crate) fn check_attr(path: &Path) -> Option<DiffAttribute> {
```
Wilfred: I think should either be `check_diff_attr`, or take an attribute name as an argument.

src/main.rs
```diff
@@ -47,6 +47,7 @@ mod diff;
 mod display;
 mod exit_codes;
 mod files;
+mod git;
```
Wilfred: Could we call this `mod gitattributes`? It's not general git functionality.

src/main.rs
```diff
@@ -414,8 +416,11 @@ fn diff_file(
     let (mut lhs_src, mut rhs_src) = match (
         guess_content(&lhs_bytes, lhs_path, binary_overrides),
         guess_content(&rhs_bytes, rhs_path, binary_overrides),
+        check_attr(Path::new(display_path)),
```
Wilfred: This covers the case when someone writes `*.ext -diff`, but not `*.ext binary`, which is also mentioned in #466.

Apparently `binary` means `-text -diff -merge` according to https://git-scm.com/docs/gitattributes#_using_macro_attributes

I think we should support `-text` and `binary` too, and also treat them as forcing binary.

Wilfred (changes requested):
Overall the approach here looks totally sensible, thanks for taking the time to write a PR!

I do think it needs a test: perhaps the easiest thing would be to write a unit test that e.g. `"package.json\0diff\0unspecified\0"` parses to the value you expect, and use that function after getting the raw string from `git check-attr`.

Otherwise, I think we should support at least the `binary` attribute too, and this is good to go. Thanks again, and let me know if you need any help :)

src/main.rs
WGH-: Yeah, that's probably a good idea, since specifying both `binary` and an external diff driver with `diff=`  is valid.

src/git.rs
WGH-: Running `git` outside of git repository is already handled by `output.status.success()` check, which gets logged with debug level.

I erroneously claimed that git being unavailable also goes there, but yeah, it actually ends up as a warning here.

Changed it to `debug`.

src/git.rs
WGH-: Done

src/git.rs
WGH-: I suppose this is no longer relevant since I changed this variant to `Other(String)`

src/git.rs
WGH-: I changed this to `check_diff_attr`. It also checks `binary` attribute under the hood, but I decided against complicating thing to the caller too much

src/main.rs
WGH-: Yes, I agree, and done

src/main.rs
WGH-: To avoid calling `git check-attr` twice, and to hide this complexity in main, I decided to make `binary` essentially force `-diff`. Please tell if you're against this implicit behaviour

WGH-: Alternatively, we could return something like `Option<(DiffAttribute, BinaryAttribute)>`. That would be more explicit, but the patterm matching expression in main is already complicated as it is, and it will become even more ridiculous.

I don't really have a strong preference here. I can implement it either way, please tell me, how :)

Wilfred (changes requested):
Almost there! The logic looks correct to me, but it's really confusing that Unset means different things for the text and binary attributes. Could you name them say AssumeText and AssumeBinary instead?

WGH-:
@Wilfred, done

I've also updated some comments since we diverged from "return `git check-attr` data as is" quite a bit now.

Wilfred:
Merged, and thank you :)

PR #878, opened 2025-09-16
Do not use jemalloc on illumos

dancrossnyc:
This is a minor portability fix.

jemalloc does not work on illumos, but fortunately, its use within difftastic is already conditional on the target environment not being msvc, so simply extend that conditional involved to also exclude illumos.


Wilfred:
Thanks 🙂

Issue #825, opened 2025-03-22
Adding protobuf (`.proto`) support


connor15mcc:
Hi folks, any interest in adding support for protobufs with something like https://crates.io/crates/tree-sitter-proto?

I have a draft of this working well and would be happy to put up a PR to upstream it.

Wilfred:
This seems like a reasonable addition for a PR :)

connor15mcc:
Wonderful, I've opened https://github.com/Wilfred/difftastic/pull/881.

I'll be largely unavailable the next few days, feel free to fixup as you see fit, otherwise I can probably take another look this weekend. Thanks for maintaining a great tool!

PR #881, opened 2025-09-30
Support `proto` diffs using `tree-sitter-proto`

connor15mcc:
I've used a build of this to great effect on large sets of proto files. This uses the upstream package https://crates.io/crates/tree-sitter-proto grammar and a corresponding set of highlight rules.

Tracking issue: https://github.com/Wilfred/difftastic/issues/825


Wilfred:
Looks good, thanks!

PR #884, opened 2025-10-02
Upgrade tree-sitter to 0.25, bump MSRV to 1.76

wetneb:
Doing this upgrade would be very useful to unvendor more parsers, because tree-sitter 0.25 supports ABIs 13, 14 and 15 at the same time, whereas 0.24 supports ABIs 13 and 14 only.

A lot of the crates.io versions of the parsers are using ABI 15, which is blocking their unvendoring.

This comes to the cost of bumping MSRV from 1.75 to 1.76, which seems acceptable to me. This is still far behind the version in Debian stable, 1.85.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,10 +129,11 @@ checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 
 [[package]]
 name = "cc"
-version = "1.2.7"
+version = "1.2.39"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Wilfred:
I'm not opposed to bumping the MSRV, but are you sure it's necessary? You've not updated the GitHub actions to use a newer rustic but the CI is green.

wetneb:
Good point, it looks like when we run `cargo test` there, it re-installs the appropriate toolchain as specified in the `rust-toolchain.toml`. But it's still worth updating the CI config so that only one toolchain gets installed, let me do that.

Wilfred:
Thanks :)

PR #886, opened 2025-10-02
Unvendor tree-sitter-hcl

wetneb:
The repo URL for the grammar has changed, but that's because the maintainer transferred their repo to the "tree-sitter-grammars" org.
I have kept the "vendored" highlights query because the grammar doesn't provide one.


Wilfred:
You even updated the manual, perfect :)

PR #888, opened 2025-10-04
Unvendor tree-sitter-r

wetneb:
Another simple case, where the crate matches the repository that had been vendored before.


Wilfred:
I'm assuming these are slightly newer versions of the parsers? If so, would you mind updating the changelog?

wetneb:
Sure, like this?

Wilfred:
Just the job :)

PR #889, opened 2025-10-04
Unvendor tree-sitter-dart

wetneb:
The dart grammar has a complicated status:
* the upstream repo, [UserNobody14/tree-sitter-dart](https://github.com/UserNobody14/tree-sitter-dart), doesn't upload to crates.io
* a [tree-sitter-dart crate](https://crates.io/crates/tree-sitter-dart) exists, but it's owned by [an abandoned fork](https://github.com/ast-grep/tree-sitter-dart) 

So for Mergiraf I have made another fork [tree-sitter-dart-orchard](https://codeberg.org/grammar-orchard/tree-sitter-dart-orchard), which tracks the upstream repository and uploads it to crates.io. So I am proposing to use that here too.


Wilfred:
Thanks.

Issue #891, opened 2025-10-06
Unvendoring parsers


wetneb:
This is a tracking issue for our initiative to unvendor as many parsers as possible (to benefit from grammar updates, help with #343 and enable Debian packaging).

Feel free to update this table as the status of crates change, and remove rows when parsers get unvendored.

| Grammar name | Vendored repo | On crates.io  | Notes |
|-------------------------|-----------------------|-----------------------------|----------|
| tree-sitter-ada | [briot/tree-sitter-ada](https://github.com/briot/tree-sitter-ada) | ❎ [requested](https://github.com/briot/tree-sitter-ada/issues/15) |  |
| tree-sitter-clojure | [sogaiu/tree-sitter-clojure](https://github.com/sogaiu/tree-sitter-clojure)| [tree-sitter-clojure](https://crates.io/crates/tree-sitter-clojure) (third party upload) | Requires Rust edition 2024, [downgrade requested](https://github.com/jerielverissimo/tree-sitter-clojure/pull/1) |
| tree-sitter-commonlisp | [tree-sitter-grammars/tree-sitter-commonlisp](https://github.com/tree-sitter-grammars/tree-sitter-commonlisp) | [tree-sitter-commonlisp](https://crates.io/crates/tree-sitter-commonlisp) | Requires Rust edition 2024 |
| tree-sitter-devicetree | [joelspadin/tree-sitter-devicetree](https://github.com/joelspadin/tree-sitter-devicetree)  | [tree-sitter-devicetree](https://crates.io/crates/tree-sitter-devicetree) | ✅ looks good |
| tree-sitter-elisp | [wilfred/tree-sitter-elisp](https://github.com/Wilfred/tree-sitter-elisp) | [tree-sitter-elisp](https://crates.io/crates/tree-sitter-elisp) | Needs updating to TS 0.25, [requested](https://github.com/Wilfred/tree-sitter-elisp/issues/8) |
| tree-sitter-elm |[elm-tooling/tree-sitter-elm](https://github.com/elm-tooling/tree-sitter-elm)   | [tree-sitter-elm](https://crates.io/crates/tree-sitter-elm) | ✅ looks good |
| tree-sitter-elvish |  [ckafi/tree-sitter-elvish](https://github.com/ckafi/tree-sitter-elvish)| ❎ |  |
[...]

PR #892, opened 2025-10-06
Unvendor tree-sitter-devicetree

wetneb:
For #891.


sample_files/compare.expected
```diff
@@ -56,7 +56,7 @@ sample_files/dart_1.dart sample_files/dart_2.dart
 107d073c2922ff14e9ba66ee0f489592  -
 
 sample_files/devicetree_1.dts sample_files/devicetree_2.dts
-f7c4e7b4444b02d87b2eec1485d86211  -
```
wetneb, line 59: There is a small change in highlighting output, with one `/` character becoming bold in the new version in the sample file. This looks acceptable to me. The highlighting for this language remains very minimalistic both before and after the unvendoring.

Wilfred: Fair enough. I use this file to see if anything has changed, but it's often minor things that are totally reasonable.

PR #901, opened 2025-10-12
Unvendor tree-sitter-ada

wetneb:
For #891.


wetneb:
Makes sense, will do. It's going to make Debian packaging easier too.

PR #904, opened 2025-10-13
Add tree-sitter-gdscript and tree-sitter-godot for Godot development

sockeye-d:
I'm not sure if I did this right, but I added two new tree-sitter definitions for GDScript files (`*.gd`) and Godot text resources (`*.tres`, `*.tscn`, `project.godot`).

Neither of them have highlighting — I'm not sure if there's a scm file for GDScript already, and I don't know how to make one.


wetneb:
I'm not a maintainer, but be aware that we're in the process of unvendoring parsers so that they are fetched from crates.io instead (#891), so I would be reluctant to introduce new parsers that are vendored (I think the godot ones are available on crates.io).

I updated the recommended process here: https://github.com/Wilfred/difftastic/pull/902

That being said, I think given that there are quite a few open PRs adding support for new languages, it's probably worth checking if new languages are welcome at the moment or not (before you put more work into this).

PR #905, opened 2025-10-14
Migrate to tree-sitter-sequel

wetneb:
This PR changes the SQL grammar from https://github.com/m-novikov/tree-sitter-sql to https://github.com/derekstride/tree-sitter-sql, actively maintained, uploaded to crates.io and providing a highlights query.

It adds sample SQL files as there were none so far.

For #891.


Wilfred:
Thanks, this seems like a nice solution. The previous SQL parser had the biggest parser.c by a significant margin, so it's nice finding an alternative.

PR #907, opened 2025-10-22
Fix typo in installation documentation

moreal:
There was a typo in the `manual/src/installation.md` file, where "Diffstatic" was written instead of "Difftastic". Looking at the history, it was originally "Difftastic" but changed to "Diffstatic" during the modification process. Although minor, I thought this could cause some confusion when reading the installation document, so I'm submitting a PR. 🙏🏻


Wilfred:
Nice catch, thanks!

PR #915, opened 2025-11-06
Unvendor tree-sitter-clojure

wetneb:
For this parser, the upstream author has made clear [their intention not to have bindings in their repository](https://github.com/sogaiu/tree-sitter-clojure/issues/64) (so, a fortiori, not to upload them to crates.io). Another repository had done so, but it requires a too recent version of Rust for us, and they haven't been responsive to [my request to change that](https://github.com/jerielverissimo/tree-sitter-clojure/pull/1).

So I have made a fork which adds those bindings and uploads them to crates.io. Like all my grammar forks, it is stored in https://codeberg.org/grammar-orchard, which is set up with the intention to enable others to also participate in grammar development and maintenance under the same roof, without the need to fork.

For #891.


Wilfred:
Huh, it's a shame upstream doesn't want to package it. The grammar is changing slowly though (last parsing change was 2023 if I'm reading the changelog correctly) so this is probably a reasonable tradeoff.

PR #923, opened 2025-11-19
Add metadata to allow fast create .deb package with "cargo deb" command

iav:
With this metadata it's easy to create .deb file easy — just `cargo deb`.


Wilfred:
Seems reasonable, thanks.

PR #930, opened 2025-12-07
docs(installation): add mise alternative method installation

jylenhof:


Wilfred:
Huh, mise is new to me. It looks good, but maybe it should be lower on the page? I think the existing installation options should probably come first.

PR #940, opened 2026-01-05
Update to tree-sitter-swift 0.7.1

ValentinWalter:
Updates tree-sitter-swift to parse the latest language updates (most notably, trailing closures in parameter lists).


Wilfred:
Thanks :)

PR #943, opened 2026-01-24
don't use jemalloc for Windows targets

ognevny:
I tried to add difftastic package in https://github.com/msys2/MINGW-packages/pull/27535, but I didn't manage to get successful build of jemalloc-sys on windows-gnu. I'll try to investigate it later, for now I suggest to just not use jemalloc for all Windows targets


Wilfred:
Seems reasonable, thanks. All the allocators have been really tricky to use when I'm trying to support all the major platforms. FWIW jemalloc's docs do suggest limiting to msvc, but I've no objection to covering all Windows. I'd rather the build works even if it costs some runtime performance.

PR #949, opened 2026-02-06
Documentation fixes

anton-ryzhov:
`dlog` was not different from the standard `log` command.
"Suggest an edit" button was targeting to 404

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,7 +129,7 @@ tree-sitter-xml = "0.7.0"
 tree-sitter-yaml = "0.7.0"
 tree-sitter-zig = "1.1.2"
 
-[target.'cfg(not(any(target_env = "msvc", target_os = "illumos", target_os = "freebsd")))'.dependencies]
+[target.'cfg(not(any(windows, target_os = "illumos", target_os = "freebsd")))'.dependencies]
 tikv-jemallocator = "0.6"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


Wilfred:
`dlog` is different from `log`, it sets `--ext-diff`. So you can do `dlog -p` for a log with difftastic showing changes, or `log -p` for conventional diff. Clearly the docs don't explain themselves sufficiently here.

The book.toml totally makes sense though. Would you like to split up the two changes, or just revert the git.md change?

anton-ryzhov:
Yeah, but what's the purpose of `dlog` without `-p`? That doesn't show any diffs, so `difft` is not used, it's the same as standard `log`.

My understanding was that with this config a user has 3 options:
* `git log` for logs without diffs
* `git log -p` for standard patches
* `git dlog` (without `-p`) for fancy patches

That's exactly what `dl` does. I thought the only difference between `dlog` and `dl` is command's length.

Did I get it wrong?

anton-ryzhov:
I wanted to split the PR now and noticed it was already done recently.

What about the remaining changes?

Wilfred:
I've clarified the comment to add `-p` to the example invocation, see 8997b527988e17b43e0dd5f3ae14444ba8d01a8f. The general intent is that `git dfoo` is equivalent to `git foo` with difftastic configured, but if that git invocation doesn't show diffs then difftastic won't be invoked.

Thanks for the PR and discussion, I'm sure you weren't the only person confused by this! :)

PR #956, opened 2026-03-07
fix: don't count javascript template strings as atoms

paperclover:
if template strings are atoms, changing the indentation of a block of code containing one counts that as a change, which is bad. instead, recurse the expression as normal.

not very happy with adding a special case to `atom_from_cursor` but i'm not familiar enough to know the right strategy.

test case
```ts
let x = `hello ${y
  ? 1
  : 2
}`;
```
```ts
{
  let x = `hello ${y
    ? 1
    : 2
  }`;
}
```

### DIFF
--- a/src/parse/tree_sitter_parser.rs
+++ b/src/parse/tree_sitter_parser.rs
@@ -593,7 +593,7 @@ pub(crate) fn from_language(language: guess::Language) -> TreeSitterConfig {
 
             TreeSitterConfig {
                 language: language.clone(),
-                atom_nodes: ["string", "template_string", "regex"].into_iter().collect(),
+                atom_nodes: ["string", "regex"].into_iter().collect(),
                 delimiter_tokens: vec![
                     ("[", "]"),
                     ("(", ")"),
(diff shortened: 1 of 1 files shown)


hmknhovgd-boop:
> إذا كانت سلاسل القوالب عبارة عن ذرات، فإن تغيير مسافة بادئة لكتلة من التعليمات البرمجية تحتوي على واحدة منها يُعتبر تغييرًا، وهذا غير مقبول. بدلاً من ذلك، قم بتكرار التعبير كالمعتاد.
> 
> لستُ سعيدًا جدًا بإضافة حالة خاصة، `atom_from_cursor`لكنني لستُ على دراية كافية بمعرفة الاستراتيجية الصحيحة.
> 
> حالة اختبار
> 
> ```ts
> let x = `hello ${y
>   ? 1
>   : 2
> }`;
> ```
> 
> ```ts
> {
>   let x = `hello ${y
>     ? 1
>     : 2
>   }`;
> }
> ```

#954

Wilfred:
This usually doesn't work, unfortunately, which is why strings with interpolated expressions are treated as atoms in most cases.

The problem is that the text before and after the interpolation needs to be present in the AST, or difftastic won't understand that the following string literals are different.

```
`foo${name}`;
`foo ${name} `;
```

paperclover:
> The problem is that the text before and after the interpolation needs to be present in the AST, or difftastic won't understand that the following string literals are different.

works in the diff

CleanShot 2026-03-20 at 21 10 38@2x

though it does make whitespace only changes a bit harder to parse
CleanShot 2026-03-20 at 21 14 05@2x

actually, this one is tough. the underlines no longer appear

CleanShot 2026-03-20 at 21 15 27@2x

i do wonder whats a reasonable solution to this. the indentation showing up as a changed node is very unsatisfying.

paperclover:
closing as stale on my part, not really sure of a good solution.

PR #960, opened 2026-03-12
Fix inline display ordering: deleted lines appeared before unrelated additions

ruudk:
### Fix inline display ordering: deleted lines appeared before unrelated additions

Given this diff between two PHP files:

```diff
 namespace App\Service;

+use App\Model\Recipient;
 use App\Context\ServiceContext;

 interface Handler
 {
+    /**
+     * @param list<int|Recipient> $recipients
+     */
     public function handle(
         ServiceContext $context,
-        int ...$ids,
+        array $recipients,
     ) : void;
 }
```

The inline display mode rendered the deleted line too early, before
the added use import and docblock:

```
5    namespace App\Service;
13            int ...$ids,          <- deleted, shown too early
   7 use App\Model\Recipient;
   12     /**
   13      * @param ...
   14      */
   17         array $recipients,    <- replacement, far below
```

After this fix, deletions appear adjacent to their replacements:

```
5    namespace App\Service;
   7 use App\Model\Recipient;
   12     /**
   13      * @param ...
   14      */
13            int ...$ids,          <- now right before replacement
   17         array $recipients,
```

The root cause was that the inline display iterated hunk lines in two
separate passes: first all old (deleted) lines, then all new (added)
lines. The fix merges these into a single loop that processes both
sides of each matched pair together.


ruudk:
@Wilfred This one is ready whenever you are :) Thanks for making diffs fantastic 😍

PR #968, opened 2026-03-31
Add AL (Business Central) language support

SShadowS:
## Summary
- Add tree-sitter-al parser (crate v2.3.0 from crates.io) for AL language support
- AL is the programming language used for Microsoft Dynamics 365 Business Central
- Detects `.al` files, includes syntax highlighting via the crate's built-in HIGHLIGHTS_QUERY

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -276,6 +276,7 @@ dependencies = [
  "tikv-jemallocator",
  "tree-sitter",
  "tree-sitter-ada",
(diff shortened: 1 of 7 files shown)


Wilfred:
Thanks for the PR! I'm afraid the difftastic binary is getting really big, so I'm being pretty selective of languages that get added at this point. I do eventually hope to support external parsers to unblock use cases like these.

PR #971, opened 2026-04-14
feat: add asm support

Neptunium931:
Added assembly using [RubixDev/tree-sitter-asm](https://github.com/RubixDev/tree-sitter-asm)


Cargo.toml
```diff
@@ -129,6 +129,7 @@ tree-sitter-verilog = "1.0.3"
 tree-sitter-xml = "0.7.0"
 tree-sitter-yaml = "0.7.0"
 tree-sitter-zig = "1.1.2"
+tree-sitter-asm = { git = "https://github.com/RubixDev/tree-sitter-asm", branch = "main", version = "0.24.0" }
```
Wilfred, line 132: This is available directly on crates.io, please depend on it there rather than referencing the git repo directly.

Wilfred (changes requested):
One small issue, and this needs rebasing I'm afraid, but otherwise looks perfect!

Wilfred:
Incidentally, have you used it much? Assembly has very little syntactic structure so I'm wondering how much difftastic helps here.

Cargo.toml
Neptunium931: Hello, I chose to use Git instead of crates.io because the latter is not up to date.
If you prefer to use only crates.io, I will still make the change.

Neptunium931:
Yes, I tested them on projects to see the results, and I found that it allows for a much better visualization of modifications to instructions, operands, or labels.
Above are two examples that highlight the difference.
image

image
Note: that this second screenshot was taken with this PR RubixDev/tree-sitter-asm/pull/28 that I submitted and for which I am awaiting a response.

Cargo.toml
Wilfred: I'd prefer crates.io regardless but I do see that 0.24 is published: https://crates.io/crates/tree-sitter-asm/versions

Wilfred (approved):
Looks great! Just needs rebasing so I can merge it :)

Neptunium931:
> Looks great! Just needs rebasing so I can merge it :)

Done

Wilfred:
Thanks :)

PR #980, opened 2026-05-10
Show unchanged lines that fall inside an inline hunk

### LINKED ISSUE #978: Incorrect diff with git config
Baseline - `delta 0.19.2`:
Image

Difftastic seems to be ignoring headings in git config.
Specifically, it vanishes this text: `[url "ssh://git@github.com"]`

`Difftastic 0.69.0  Revision:  90a0f1b6a 2026-04-29`
Image

ChrisJr404:
Closes #978.

## Bug

When two novel lines in a hunk are close enough that the merge step in `display::hunks::merge_adjacent` combines them into one hunk, the unchanged lines between them are silently dropped from inline (`--display=inline`) output. Side-by-side and JSON output don't have this problem because both use `matched_lines_indexes_for_hunk` to walk the full hunk range, so they already include the in-between unchanged lines. Inline was iterating `hunk.lines` directly, and that field only contains positions with novel content.

In the report this surfaces as a missing section heading on a `.gitconfig`:

```
$ difft before.gitconfig after.gitconfig --display=inline
```

The line `[url "ssh://git@github.com"]` between two changed `pushInsteadOf =` lines does not appear.

## Repro

`before.gitconfig`:

```
[uploadpack]
    allowFilter = true

[url "ssh://aur@aur.archlinux.org"]
    pushInsteadOf = "https://aur.archlinux.org"

[url "ssh://git@github.com"]
    pushInsteadOf = "https://github.com"

[versionsort]
    # prereleaseSuffix is deprecated in favour of suffix
```

`after.gitconfig` (same, but both `pushInsteadOf` lines are commented out):

```
[uploadpack]
    allowFilter = true

[url "ssh://aur@aur.archlinux.org"]
    ; pushInsteadOf = "https://aur.archlinux.org"

[url "ssh://git@github.com"]
    ; pushInsteadOf = "https://github.com"

[versionsort]
    # prereleaseSuffix is deprecated in favour of suffix
```

### Before this PR

```
/tmp/after3.gitconfig --- Text
1    [uploadpack]
2        allowFilter = true
3    
4    [url "ssh://aur@aur.archlinux.org"]
5        pushInsteadOf = "https://aur.archlinux.org"
8        pushInsteadOf = "https://github.com"
   5     ; pushInsteadOf = "https://aur.archlinux.org"
   8     ; pushInsteadOf = "https://github.com"
   9 
   10 [versionsort]
   11     # prereleaseSuffix is deprecated in favour of suffix
```

Lines 6 (blank) and 7 (`[url "ssh://git@github.com"]`) are missing on both sides.

### After this PR

```
/tmp/after3.gitconfig --- Text
1    [uploadpack]
2        allowFilter = true
3    
4    [url "ssh://aur@aur.archlinux.org"]
5        pushInsteadOf = "https://aur.archlinux.org"
6    
7    [url "ssh://git@github.com"]
8        pushInsteadOf = "https://github.com"
   5     ; pushInsteadOf = "https://aur.archlinux.org"
   6 
   7 [url "ssh://git@github.com"]
   8     ; pushInsteadOf = "https://github.com"
   9 
   10 [versionsort]
   11     # prereleaseSuffix is deprecated in favour of suffix
```

## Fix

Add a small `fill_in_hunk_gaps` helper in `src/display/inline.rs` that walks `hunk.lines` and emits the intervening line numbers for each side, then update the LHS and RHS print loops to consult `hunk.novel_lhs` / `hunk.novel_rhs` so the new gap entries render with dimmed (context) line numbers instead of bold red / green ones.

I considered switching inline to use `all_matched_lines_filled` + `matched_lines_indexes_for_hunk` the same way side-by-side does, but that would also require restructuring how the before / after context loops are integrated. This narrower change keeps the existing structure and only fixes the dropped-context-inside-hunk case.

## Testing

- Added four unit tests for `fill_in_hunk_gaps` covering the empty input, no-gap, LHS-only-changes-with-an-unchanged-tail, and unequal-side-gaps shapes.
- `cargo test --release` (148 tests) passes.
- Manually verified against the `.gitconfig` repro above and against syntactic (Rust) inputs to confirm the multi-hunk path still splits correctly when changes are far apart.

### DIFF
--- a/src/display/inline.rs
+++ b/src/display/inline.rs
@@ -1,5 +1,7 @@
 //! Inline, or "unified" diff display.
 
+use line_numbers::LineNumber;
+
 use crate::constants::Side;
 use crate::display::context::{
     calculate_after_context, calculate_before_context, opposite_positions,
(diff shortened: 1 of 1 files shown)


src/display/inline.rs
```diff
@@ -11,6 +13,50 @@ use crate::options::DisplayOptions;
 use crate::parse::syntax::MatchedPos;
 use crate::summary::FileFormat;
 
+/// Expand `hunk.lines` to include any unchanged lines that fall between
+/// changed lines on the same side. Without this, inline display drops
```
Wilfred, line 17: Just keep the first sentence: the rest is details of the bug which you've just fixed.

src/display/inline.rs
```diff
@@ -156,3 +204,92 @@ pub(crate) fn print(
         println!();
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Wilfred, line 209: Don't add unit test here: instead add a CLI integration test that verifies the missing line is now shown.

Wilfred (changes requested):
Thanks for the PR! This a definitely a good direction, just needs a few fixes (and a rebase) before it's good to go 😊

PR #988, opened 2026-06-08
Add Fish parser support

### LINKED ISSUE #987: Missing Fish shell language support
Fish shell scripts currently fall back to plain text diffing.

There is already an open PR adding Fish support: https://github.com/Wilfred/difftastic/pull/847 But it's kinda outdated, etc.

So I opened a newer PR based on current `master` with parser configuration, docs, and regression samples: https://github.com/Wilfred/difftastic/pull/988

`tree-sitter-fish` exists and is the parser used by both PRs:
https://github.com/ram02z/tree-sitter-fish

escape0707:
## Summary

Adds Fish shell support using `tree-sitter-fish`. This includes language detection for `*.fish`, Emacs mode headers, and Fish shebangs, plus parser configuration, supported-language documentation, and regression samples.

Fixes #987.

## Validation

- `cargo test`
- `sample_files/compare_all.sh`


src/parse/tree_sitter_parser.rs
```diff
@@ -411,6 +411,22 @@ pub(crate) fn from_language(language: guess::Language) -> TreeSitterConfig {
                 sub_languages: vec![],
             }
         }
+        Fish => {
+            let language = tree_sitter_fish::language();
```
edouard-lopez, line 432: question: Fish is the only parser to [return a function](https://github.com/ram02z/tree-sitter-fish/blob/f435b0bd772578c70e5d158b85267bb886316f88/bindings/rust/lib.rs#L27-L29), while other return a constant (e.g. [bash](https://github.com/tree-sitter/tree-sitter-bash/blob/a06c2e4415e9bc0346c6b86d401879ffb44058f7/bindings/rust/lib.rs#L32-L33)). What is the impact?

edouard-lopez (commented):
Looking good to me. However, I'm not familiar with neither tree-sitter nor diffstatic code 😅

src/parse/tree_sitter_parser.rs
escape0707: Good question. This is just a difference in the Rust bindings exposed by `tree-sitter-fish`: `tree_sitter_fish::language()` already returns a `tree_sitter::Language`, while newer crates like Bash expose a `LanguageFn` constant that Difftastic wraps with `tree_sitter::Language::new(...)`. Difftastic stores the result as the same `tree_sitter::Language` either way and later passes it to `Parser::set_language`, so I do not expect any behavioral impact here.

Wilfred:
This looks great, thanks for the PR :)

PR #997, opened 2026-06-18
Use a consistent file path label when paths share no suffix

### LINKED ISSUE #925: Inconsistent choice for shown file path
Hi!

When diffing two files *without* file extension, the first one is chosen to label the hunk. However, when the files have got an extension, the second one is chosen instead. In the following screenshot, observe that in the first invocation `a` is chosen, in the second `b.txt`.
In general, perhaps difft should detect a rename in this case and display both the left and right path, instead of picking a single one?

Image

[a.txt](https://github.com/user-attachments/files/23687972/a.txt)
[b.txt](https://github.com/user-attachments/files/23687974/b.txt)

`difft --version`
```
Difftastic 0.67.0

Toolchain: 1.92.0
System:    linux x86_64

```

StressTestor:
## what

When diffing two files whose paths share no common trailing component, the path shown in the hunk header depends on whether the files have an extension:

```
$ difft a b
a --- Text
$ difft a.txt b.txt
b.txt --- Text
```

`a` vs `b` labels with the left path, but `a.txt` vs `b.txt` labels with the right path. This is #925.

## why

`build_display_path`'s no-common-suffix arm picked the rhs path when rhs had an extension and the lhs path otherwise:

```rust
None => {
    if rhs.extension().is_some() {
        rhs.display().to_string()
    } else {
        lhs.display().to_string()
    }
}
```

so the side shown was driven by the extension.

## fix

Use the rhs path consistently here, matching the git-tmpfile branch just above (which already returns rhs) and the usual convention of showing the new/destination file.

```rust
None => rhs.display().to_string(),
```

The change is scoped to that one match arm. The reporter also suggested detecting a rename and showing both paths; that is a larger behaviour change, so I kept this minimal and easy to extend if you would rather take the show-both approach.

## tests

`test_build_display_path_consistent_without_common_suffix` covers the extensionless and extensioned cases (both now use rhs) and the shared-suffix case (unchanged). It fails before the fix and passes after. Full suite green.

Fixes #925.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ Improved diffing performance, particularly when diffing directories.
 
 Removed support for SCSS (upstream parser is no longer maintained).
 
+### Display
+
+Fixed an inconsistent choice of file path label when diffing two paths
+that share no common suffix: the label no longer depends on whether the
+files have an extension.
+
 ## 0.69 (released 30th April 2026)
 
 ## Diffing
--- a/src/options.rs
+++ b/src/options.rs
@@ -620,13 +620,10 @@ fn build_display_path(lhs_path: &FileArgument, rhs_path: &FileArgument) -> Strin
 
             match common_path_suffix(lhs, rhs) {
                 Some(common_suffix) => common_suffix,
-                None => {
-                    if rhs.extension().is_some() {
-                        rhs.display().to_string()
-                    } else {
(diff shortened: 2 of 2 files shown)


Wilfred:
Thanks for the PR! I ended up merging a separate fix.

Please also note that the project has an AI_POLICY.md since you created this PR.

PR #1000, opened 2026-06-24
Don't emit hunks with no novel lines

### LINKED ISSUE #995: Crashes on python file with "Hunk lines should be present in matched lines"
(1) A description of the issue. A screenshot is often helpful too.
```
% make diff
difft old.py new.py

thread 'main' (17348282) panicked at /Users/illiadenysenko/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/difftastic-0.68.0/src/display/hunks.rs:667:31:
Hunk lines should be present in matched lines
make: *** [diff] Error 101
```

(2) A copy of what you're diffing. If you're diffing files, include the before and after files. If you're using difftastic with a VCS repository (e.g. git), include the URL and commit hash.
`new.py`
```python
from pathlib import Path

from fastapi import FastAPI


def create_app_from_runtime_config() -> FastAPI:
    config = load_runtime_config()
    store = RepoMarkStore.open(Path(config.db_path))

    marks = store.list()
    if len(marks) == 0:
        raise SystemExit(
            "No marked repos. Run: dirdiff mark /path/to/repo "
            "--db-path /path/to/db.sqlite"
        )
    return create_app(store, presets_root=config.presets_root)


def handle_mark_command(args: argparse.Namespace) -> None:
    store = RepoMarkStore.open(Path(args.db_path))
    repo_path = Path(args.repo_path)
    if args.name is None:
        resolved_path = repo_path.expanduser().resolve()
        display_name = resolved_path.name
        if not display_name:
            raise SystemExit(
                f"Cannot derive a repo name from path: {resolved_path}"
            )
    else:
        display_name = args.name
    try:
        mark = store.mark(RepoMarkInput(path=repo_path, name=display_name))
    except RepoRegistryError as exc:
        raise SystemExit(str(exc)) from exc
    print(f"Marked repo {mark.id}: {mark.path}", file=sys.stderr)
```

And `old.py`
```python
from pathlib import Path
from typing import Any


def create_app_from_runtime_config() -> Any:
    config = load_runtime_config()
    repo_root = (
        Path(config.repo_root).expanduser() if config.repo_root else None
    )
    repo = GitBackend.discover(repo_root=repo_root)
    presets_root = (
        Path(config.presets_root).expanduser() if config.presets_root else None
    )
    preset_repo = PresetBackend.discover(presets_root=presets_root)
    service = TextDiffService(repo)
    git_service = GitDiffService(repo)
    difftastic_service = DifftasticDiffService(repo)
    preset_service = TextDiffService(preset_repo)
    preset_git_service = GitDiffService(preset_repo)
    preset_difftastic_service = DifftasticDiffService(preset_repo)
    defaults = build_defaults(
        service,
        left=config.left,
        right=config.right,
        base_branch=config.base_branch,
        review_branch=config.review_branch,
    )
    return create_app(
        service,
        defaults,
        services={"git": git_service, "difftastic": difftastic_service},
        preset_services={
            "dirdiff": preset_service,
            "git": preset_git_service,
            "difftastic": preset_difftastic_service,
        },
    )
```

(3) The version of difftastic you're using (see `difft --version`) and your operating system.
```
% difft --version
Difftastic 0.68.0

Toolchain: 1.91.0
System:    macos aarch64
```

truffle-dev:
Fixes #995.

`difft old.py new.py` on the files in the report panics:

```
thread 'main' panicked at src/display/hunks.rs:667:31:
Hunk lines should be present in matched lines
```

### What's happening

`matched_novel_lines` pairs each novel line with the matched line opposite it, e.g. `(Some(novel_lhs), Some(opposite_rhs))`. `lines_to_hunks` then runs the pairs through `enforce_increasing`, which replaces a line number with `None` when it occurs out of order. When the novel side of a pair is the one discarded, only the matched (non-novel) line is left behind, e.g. `(None, Some(opposite_rhs))`.

A run of these stripped pairs can form an entire hunk. `find_novel_lines` finds nothing novel in it, so the hunk's `novel_lhs` and `novel_rhs` are both empty. `matched_lines_indexes_for_hunk` then searches `matched_lines` for the hunk's smallest and largest novel line, gets `(None, None)` for both, and `either_side_equal` is never true, so the `expect` at hunks.rs:667 fires.

I confirmed this against the panic site by printing the hunk and matched lines:

```
hunk_smallest=(None, None) hunk_largest=(None, None)
novel_lhs_sorted=[] novel_rhs_sorted=[]
start_i=None end_i=None
```

### The fix

A hunk with no novel lines represents no change, so it shouldn't be produced in the first place. This skips emitting a hunk when both novel sets are empty, which keeps the invariant that `matched_lines_indexes_for_hunk` already relies on.

### Test

Added `empty_novel_hunk_doesnt_crash` to `tests/cli.rs`, using the two files from the report as fixtures. It panics on `master` and passes with this change. The existing unit and CLI tests still pass.

Thanks to @juliancoffee for the clear repro; their fork had a similar observation about the empty-hunk case.

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -137,6 +137,21 @@ fn text_changes_at_end_doesnt_crash() {
     cmd.assert().success();
 }
 
+/// Regression test for https://github.com/Wilfred/difftastic/issues/995.
+///
+/// When line reordering strips the novel side from every line pair in a
+/// group, the resulting hunk has no novel lines, which used to panic with
+/// "Hunk lines should be present in matched lines".
+#[test]
+fn empty_novel_hunk_doesnt_crash() {
+    let mut cmd = get_base_command();
+
+    cmd.arg("sample_files/cli_tests/empty_hunk_1.py")
+        .arg("sample_files/cli_tests/empty_hunk_2.py");
+
+    cmd.assert().success();
+}
+
 #[test]
 fn makefile_text_as_atom() {
     let mut cmd = get_base_command();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 Improved diffing performance, particularly when diffing directories.
 
+Fixed a crash when line reordering left a hunk with no novel lines
+(#995).
+
 ### Parsing
 
 Removed support for SCSS (upstream parser is no longer maintained).
--- a/sample_files/cli_tests/empty_hunk_1.py
+++ b/sample_files/cli_tests/empty_hunk_1.py
@@ -1,0 +1,37 @@
+from pathlib import Path
+from typing import Any
+
+
+def create_app_from_runtime_config() -> Any:
+    config = load_runtime_config()
+    repo_root = (
+        Path(config.repo_root).expanduser() if config.repo_root else None
+    )
+    repo = GitBackend.discover(repo_root=repo_root)
+    presets_root = (
+        Path(config.presets_root).expanduser() if config.presets_root else None
+    )
+    preset_repo = PresetBackend.discover(presets_root=presets_root)
+    service = TextDiffService(repo)
+    git_service = GitDiffService(repo)
+    difftastic_service = DifftasticDiffService(repo)
+    preset_service = TextDiffService(preset_repo)
+    preset_git_service = GitDiffService(preset_repo)
+    preset_difftastic_service = DifftasticDiffService(preset_repo)
+    defaults = build_defaults(
+        service,
+        left=config.left,
+        right=config.right,
+        base_branch=config.base_branch,
+        review_branch=config.review_branch,
+    )
+    return create_app(
+        service,
+        defaults,
+        services={"git": git_service, "difftastic": difftastic_service},
+        preset_services={
+            "dirdiff": preset_service,
+            "git": preset_git_service,
+            "difftastic": preset_difftastic_service,
+        },
+    )
--- a/sample_files/cli_tests/empty_hunk_2.py
+++ b/sample_files/cli_tests/empty_hunk_2.py
@@ -1,0 +1,35 @@
+from pathlib import Path
+
+from fastapi import FastAPI
+
+
+def create_app_from_runtime_config() -> FastAPI:
+    config = load_runtime_config()
+    store = RepoMarkStore.open(Path(config.db_path))
+
+    marks = store.list()
+    if len(marks) == 0:
+        raise SystemExit(
+            "No marked repos. Run: dirdiff mark /path/to/repo "
+            "--db-path /path/to/db.sqlite"
+        )
+    return create_app(store, presets_root=config.presets_root)
+
+
+def handle_mark_command(args: argparse.Namespace) -> None:
+    store = RepoMarkStore.open(Path(args.db_path))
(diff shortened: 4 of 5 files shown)


juliancoffee:
@truffle-dev, so what does it show on my example? Could you make a screenshot?

truffle-dev:
Here's `difft old.py new.py` on your files from #995, built from this PR:

difft on the #995 files

Exits 0, renders the diff normally. The empty novel hunk that used to be emitted between the two real hunks is just gone.

For contrast, reverting only the `hunks.rs` change and rebuilding panics on the exact same inputs, matching the original report:

```
thread 'main' panicked at src/display/hunks.rs:667:31:
Hunk lines should be present in matched lines
```

juliancoffee:
@truffle-dev just checked it again, it doesn't seem to be the full screenshot.
Or well, if it's a full screenshot, then you kinda lost four lines at the end.

Please verify.

linsomniac:
I'm having a similar problem, but this fix doesn't resolve it.  I built this branch and ran it against my repro files (attached) and am still getting a backtrace.

[difft-repro-new-identity-list-valid.py](https://github.com/user-attachments/files/29832577/difft-repro-new-identity-list-valid.py)
[difft-repro-old-identity-list-valid.py](https://github.com/user-attachments/files/29832584/difft-repro-old-identity-list-valid.py)

I had Codex build a fix, it did a "high-water-mark fix in match_novel_lines,  because it avoids generating redundant/out-of-order line pairs earlier."  I can't speak intelligently about the codebase or the fix, I just wanted to go one step further than replying with a "I'm still having problems with this fix applied" response.

Codex's fix, which does solve my repro files, if it helps, is in my branch at: https://github.com/linsomniac/difftastic/tree/empty-novel-hunk-fix

Let me know if I should submit it as a PR on your truffle-dev branch or something else.

tests/cli.rs
```diff
@@ -137,6 +137,21 @@ fn text_changes_at_end_doesnt_crash() {
...
+/// group, the resulting hunk has no novel lines, which used to panic with
+/// "Hunk lines should be present in matched lines".
+#[test]
+fn empty_novel_hunk_doesnt_crash() {
+    let mut cmd = get_base_command();
+
+    cmd.arg("sample_files/cli_tests/empty_hunk_1.py")
+        .arg("sample_files/cli_tests/empty_hunk_2.py");
+
+    cmd.assert().success();
```
KSHITIZ6341, line 152: Could we strengthen this regression test to assert the rendered output as well as the exit status?

I built this branch and ran the new fixture manually:

`./target/debug/difft sample_files/cli_tests/empty_hunk_1.py sample_files/cli_tests/empty_hunk_2.py`

It exits 0, but the output appears to stop at left line 31, while `empty_hunk_1.py` still has lines 32-37 (`preset_services` and the closing lines). That seems to match the reporter's latest concern that the fix avoids the panic but may still drop trailing output.

If that output is not intentional, an assertion for the tail of the fixture, or snapshot coverage for this case, would help lock in both parts of the regression: no panic and no truncated diff.

KSHITIZ6341 (commented):
One regression-test coverage note.

PR #1012, opened 2026-07-17
README: note release date when mentioning v0.50

Zearin:
(Yes, I’m human. I hate that I feel the need to state that. :P)


Wilfred:
Thanks :)

PR #1013, opened 2026-07-17
Switch to tree-sitter-java-orchard

wetneb:
The [tree-sitter-java](https://github.com/tree-sitter/tree-sitter-java) repo suffers from the same sort of abandonment as other repos in this org, meaning that it is difficult to contribute improvements to it.

The [fork](https://codeberg.org/grammar-orchard/tree-sitter-java-orchard) we made in the [grammar-orchard](https://codeberg.org/grammar-orchard) org got some nice improvements, primarily around newer features of the language, so I propose to switch to it, as it should reduce the number of parse errors.


Wilfred:
It's a shame the first party parsers aren't being actively worked on.

Anyway, thanks for the PR and the upstream maintenance :)

PR #1021, opened 2026-08-01
Account for ASCII control character display width

### LINKED ISSUE #569: Unprintable characters brake side-by-side view
## A description of the issue. A screenshot is often helpful too.

I'm using `difftastic` for diffing in git with:

```
[diff]
    external = difft
```

When diffing a line with unprintable characters (like Unit Separator - 0x1F) the width of the line is not properly calculated. I think because those characters are usually displayed as two regular characters (Unit Separator is displayed as `^_`). If the lines are close to half of the screen, the double width characters make left half bigger than half of the screen, which results in the right side wrapping to to left half.

image

When it is done through `git difftool` or by comparing two files with `difft` directly, the unprintable characters are not displayed at all. I need to pipe the output to `less` to have the same effect.

## A copy of what you're diffing.

Before: `LOREM IPSUM DOLOR SIT AMET, CONSECTETUR ADIPISICING ELIT, SED DO EIUSMODTEMPOR`
After: `lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmodtempor`

you might have to add or remove some characters to make it close to half the screen.

## The version of difftastic you're using (see difft --version) and your operating system.

```
$ difft --version
Difftastic 0.51.1 (8752af74d 2023-08-25)
$ uname -a  
6.4.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 24 Aug 2023 00:38:14 +0000 x86_64 GNU/Linux
```

subotac:
## Summary
  - account for the rendered width of non-tab control characters
  - keep wrapped and side-by-side output aligned
  - add focused regression coverage

  Fixes #569

  ## Testing
  - focused regression test
  - unit and documentation tests
  - cargo fmt --all -- --check

  AI-assisted tooling was used for test development I reviewed and verified every submitted line and can explain the change.

### DIFF
--- a/src/display/style.rs
+++ b/src/display/style.rs
@@ -133,8 +133,8 @@ fn split_string_by_width(s: &str, max_width: usize, tab_width: usize) -> Vec<(&s
     parts
 }
 
-/// Return a copy of `src` with tabs expanded and ASCII control
-/// characters replaced by caret notation.
+/// Return a copy of `src` with tabs expanded and non-formatting ASCII
+/// control characters replaced by caret notation.
 pub(crate) fn replace_tabs(src: &str, tab_width: usize) -> String {
     let tab_as_spaces = " ".repeat(tab_width);
     let mut result = String::with_capacity(src.len());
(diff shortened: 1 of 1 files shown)


Wilfred:
Thanks for the PR. Have you confirmed that the characters render at the same width both with and without less?

subotac:
Thanks! I have only verified the width calculation with the regression test so far, not the actual rendering both directly and through `less`. I'll check both modes at the same terminal width and report back.

subotac:
Confirmed both paths. The updated output renders non-formatting ASCII controls as caret notation; the issue fixtures contain no raw Unit Separator bytes, and the direct output is byte-for-byte identical after `less -F -X`. All CI checks pass.

PR #1030, opened 2026-08-09
fix(cli): reject --width below the spacer floor instead of panicking

YuriNachos:
## Summary
`difft --width 2` (and `--width 0`/`1`, or `DFT_WIDTH=2`) panicked in release builds at `src/display/side_by_side.rs`:`assert!(display_width > SPACER.len())` (`SPACER` is 2 chars), because the clamped `display_width` was not wider than the spacer.

## Root cause
`--width` / `DFT_WIDTH` parsed any `usize` with no minimum. With a value <= 2, `display_width = min(terminal_width, width_without_truncation)` landed at or below the 2-character spacer, the release `assert!` (not `debug_assert!`) fired, and the message ("Terminal total width should not overflow") was misleading — the real condition was "terminal too narrow", not arithmetic overflow.

## Fix
- Validate `--width`/`DFT_WIDTH` at CLI parse time against a minimum that guarantees a two-column layout.
- Clamp terminal auto-detection to the same floor.
- Replace the release `assert!` in `SourceDimensions` with a defensive `max(display_width, SPACER.len() + 1)` clamp plus a `debug_assert!`, so an undersized width produces a usable (clamped) layout instead of a panic.

## Tests
CLI regression tests for the rejection of too-small `--width`, plus a layout test asserting the clamp path no longer panics.

### DIFF
--- a/src/display/side_by_side.rs
+++ b/src/display/side_by_side.rs
@@ -207,9 +207,17 @@ impl SourceDimensions {
             + content_max_width;
         let display_width = min(terminal_width, width_without_truncation);
 
-        assert!(
+        // A terminal narrower than the spacer cannot fit two columns, and
+        // subtracting the spacer below would underflow. The CLI rejects
+        // `--width`/`DFT_WIDTH` below MIN_TERMINAL_WIDTH and terminal
+        // auto-detection clamps to that floor, but guard defensively here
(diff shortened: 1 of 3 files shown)


Wilfred:
Please read AI_POLICY.md

PR #1032, opened 2026-08-11
Trailing commas and atom node fixes for Scala

hugo-vrijswijk:
Bump tree-sitter-scala to 0.26.2.

Add ignore_trailing_tokens for Scala. Also drop "template_string" from the Scala atom_nodes: it isn't a node kind in tree-sitter-scala, it was copied from the JavaScript config.


Wilfred:
Looks great, thank you 😊

Issue #707, opened 2024-04-29
Crash inside eshell with memory allocation error


sunlin7:
Hi,
I tried difft inside eshell and got crash, just a simple diff two files:
> $ difft fileA fileB
memory allocation of 9223372036854775803 bytes failed
aborted

The eshell is built-in emacs-30.0.50, and difft is the `Difftastic 0.57.0 (00e837a 2024-04-01, built with rustc 1.65.0)`.

Wilfred:
Could you share the files that caused this issue?

sunlin7:
Hi @Wilfred 
I attach a.el and b.el which will lead a crash message in my local with eshell inside emacs.
[test.TGZ](https://github.com/Wilfred/difftastic/files/15172295/test.TGZ)

> $ emacs -nw -q # and start eshell
/tmp $ difft a.el b.el
b.el --- Emacs Lisp
File permissions changed from 100664 to 100600.
memory allocation of 9223372036854775804 bytes failed
aborted

bernhard-herzog:
With a debug build I get a more maningful error and backtrace:

    3: difft::display::side_by_side::SourceDimensions::new
              at ./src/display/side_by_side.rs:185:31

That's

```rust
        let lhs_total_width = (terminal_width - SPACER.len()) / 2;
```

When running in eshell, `terminal_width` is 0 and SPACER.len() is
positive, hence the overflow.

AFAICT, `terminal_width` is 0, because in `detect_display_width`
in `./src/options.rs:864` there's

```rust
    if let Ok((columns, _rows)) = crossterm::terminal::size() {
```

When running in eshell, `crossterm::terminal::size()` returns Ok, but
with colums and rows both 0.

So the cause is that crossterm cannot correctly determine the size of
the terminal in eshell. The best fix is probably to fix crossterm, but
as a workaround, difftastic could check for a 0 width here and fall back
on e.g. looking at the `COLUMNS` environment variable which is set
correctly in eshell or the default value of 80.

Wilfred:
Reported upstream: https://github.com/crossterm-rs/crossterm/issues/891

sunlin7:
Thank you all for the great works!

PR #1037, opened 2026-08-13
Replace crossterm with terminal_size

Wilfred:
We only used crossterm to detect the terminal width, and clap already depends on terminal_size for wrap_help.

This removes 15 crates: crossterm, crossterm_winapi, lock_api, mio, parking_lot, parking_lot_core, redox_syscall, scopeguard, signal-hook, signal-hook-mio, signal-hook-registry, wasi, winapi, winapi-i686-pc-windows-gnu and winapi-x86_64-pc-windows-gnu. It's the last user of the unmaintained winapi 0.3. The lockfile goes from 199 packages to 175, as the older duplicate copies of bitflags and the windows-targets 0.48 family drop out too.

terminal_size looks at stdout, stderr then stdin, whereas crossterm opened /dev/tty first. Comparing the width each build detects in a pty of a known size, they agree whenever any of the three standard streams is still a terminal, including when git pipes us to a pager (stdout is a pipe, stderr is not). They differ when all three are redirected: crossterm still found the controlling terminal, whereas we now fall back to COLUMNS and then to DEFAULT_TERMINAL_WIDTH. Keeping that case would mean opening /dev/tty ourselves.

We also lose crossterm's last-ditch fallback of shelling out to tput, which the COLUMNS fallback from #707 covers.

Windows is the one thing I couldn't check locally, so that rests on CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,12 +90,6 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
-
-[[package]]
-name = "bitflags"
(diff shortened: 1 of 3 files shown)


Wilfred:
@codex review

PR #1042, opened 2026-08-14
Underline the changed parts of a renamed path

VXNCXNX:
Fixes #963.

*Disclosure per AI_POLICY.md: written with AI assistance (Claude Code). I've read every line and am accountable for it; happy to discuss any part.*

You asked to underline the changed parts of the path in a rename header. This does that, segment by segment on `/`:

```
same dir, new name    Renamed from src/[a.ts] to src/[b.ts]
same name, new dir    Renamed from [old]/a.ts to [new]/a.ts
both changed          Renamed from [old]/[a.ts] to [new]/[b.ts]
extra segment         Renamed from scripts/backup-db.ts to [src]/scripts/backup-db.ts
```

(brackets = underlined). The last one is the case from your issue: the left side is entirely unchanged, so only the added `src` is marked.

## The one design question

`extra_info` is a plain `String`, so underlining meant either embedding codes or changing the type. I kept the `String` and embedded, gated on the already-computed `use_color`. Changing the type would have rippled through `summary.rs`, `main.rs` (~10 sites), `inline.rs`, `side_by_side.rs` and `style.rs` — far more than this issue warrants. As done, exactly one call site changed.

One subtlety worth flagging: `style.rs:569` dims the whole line, so the underline is closed with `\x1b[24m` (underline-off) rather than `\x1b[0m`, which would have killed the dim for the rest of the line. That's also why owo-colors' `Style` wasn't usable here — it emits a full reset.

`--color never` returns the plain strings untouched, verified with `cat -v` for all four shapes and pinned by a test.

```
cargo test         124 passed (unit), 23 passed (integration)
cargo fmt --check  clean
```

Tests cover the four shapes above plus the no-colour case.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -639,6 +639,58 @@ fn build_display_path(lhs_path: &FileArgument, rhs_path: &FileArgument) -> Strin
     }
 }
 
+/// Underline the `/` separated segments that differ between
+/// `old_name` and `new_name`, so it's obvious which part of the path
+/// changed.
+///
+/// Segments are compared from both ends, so renaming
+/// `scripts/foo.ts` to `src/scripts/foo.ts` only underlines `src`.
+fn underline_changed_segments(old_name: &str, new_name: &str, use_color: bool) -> (String, String) {
+    if !use_color {
+        return (old_name.to_owned(), new_name.to_owned());
+    }
+
+    let old_parts: Vec<&str> = old_name.split('/').collect();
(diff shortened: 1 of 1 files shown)


Wilfred (changes requested):
We should be diffing the segments, not just stripping prefixes and suffixes. We should at least think about Windows too.

Are you happy to write comments yourself (not LLM) and discuss this, per AI_POLICY.md?

PR #1043, opened 2026-08-15
Sort the options in --help alphabetically

VXNCXNX:
Split out of #1041, as you asked.

## What's broken

`difft --help` lists options in declaration order, so related flags are scattered and there is no way to scan for one by name.

## The fix

`next_display_order(None)` tells clap to sort them alphabetically. One line, no behaviour change beyond the help output.

## Verification

`cargo build` and `difft --help` show the options sorted. `cargo test` passes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -533,7 +533,8 @@ fn diff_conflicts_file(
     if conflict_files.num_conflicts == 0 {
         print_error(
             &format!(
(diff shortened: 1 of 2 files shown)


VXNCXNX:
Closing in favour of #1044, which is the same change branched cleanly off master.

PR #1044, opened 2026-08-15
Sort the options in --help alphabetically

VXNCXNX:
Split out of #1041, as you asked.

## What's broken

`difft --help` lists options in declaration order, so related flags are scattered and there is no way to scan for one by name.

## The fix

`next_display_order(None)` tells clap to sort them alphabetically. One line, no behaviour change beyond the help output.

## Verification

`cargo build` and `difft --help` show the options sorted. `cargo test` passes.

*Disclosure per AI_POLICY.md: written with AI assistance (Claude Code). I have read every line and am accountable for it; happy to discuss any part.*

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -139,6 +139,9 @@ fn app() -> clap::Command {
     after_help.push('.');
 
     Command::new("Difftastic")
+        // Show options in alphabetical order, rather than in
+        // declaration order.
+        .next_display_order(None)
         .override_usage(USAGE)
         .version(env!("CARGO_PKG_VERSION"))
         .long_version(VERSION.as_str())


Wilfred:
How does the order look now? Do you think it needs more grouping?

VXNCXNX:
Here is the full list as it stands:

```
--background      --list-languages
--byte-limit      --override
--check-only      --override-binary
--color           --parse-error-limit
--context         --skip-unchanged
--display         --sort-paths
--exit-code       --strip-cr
--graph-limit     --syntax-highlight
-h, --help        --tab-width
--ignore-comments -V, --version
                  --width

DEBUG OPTIONS:
--dump-syntax
--dump-syntax-dot
--dump-ts
```

`DEBUG OPTIONS` keeps its own heading, so the three dump flags stay out of the main list either way.

Two things you may or may not like:

`-h` and `-V` sort under h and V, by their short names, so `--version` lands between `--tab-width` and `--width` rather than at the end. clap does that on its own.

On grouping: I would say no for now. There are only twenty options in the main list, which is short enough to scan, and the natural clusters are small and would not divide cleanly. `--byte-limit`, `--graph-limit` and `--parse-error-limit` are all limits, but `--context` and `--width` are display concerns that some people would file under limits too. `--background`, `--color` and `--syntax-highlight` are colour, though `--display` and `--width` are arguably the same family.

If you do want headings, the split I would trust is a small OUTPUT group (`--background`, `--color`, `--display`, `--syntax-highlight`, `--tab-width`, `--width`) and leave everything else ungrouped. Happy to push that here if you like it, or leave the flat alphabetical list.

Wilfred:
Are you writing PR comments with AI?

VXNCXNX:
Yes, and I owe you a straight answer plus an apology for making you ask.

I use Claude Code. It writes the code and the prose, and I review and test before sending. Your AI_POLICY.md asks for disclosure in the PR description, and I did that on #1041 and #1042 but not on #1043 or this one. That was sloppy on my part, not a deliberate omission: when I split this out of #1041 at your request I rewrote the description and dropped the line. I have just added it back here.

On the substance of this PR, so you can judge it rather than the process: it is `next_display_order(None)` on the clap command. I built it and read `difft --help` before and after, and the options are sorted alphabetically instead of in declaration order. Nothing else changes. Ask me anything about it.

I understand if the answer changes how you want to handle these three PRs. If you would rather not take AI-assisted contributions in practice, whatever the policy allows on paper, say so and I will close them and stop, no argument. Your project, your call, and I would rather hear that than have you spend review time on things you did not want.

PR #1052, opened 2026-08-24
Fix Ruby heredocs highlighting unchanged lines

### LINKED ISSUE #994: Weird diff for ruby heredocs
Maybe this is more related to [tree-sitter-ruby](https://github.com/tree-sitter/tree-sitter-ruby/), but I'm getting these:
Difftastic:
Image
git:
Image

```
Difftastic 0.69.0

Toolchain: 1.95.0
System:    macos aarch64
```

alagos:
# Why

Fixes #994.

`MatchedPos::new` trims zero-width spans from the ends of an atom's position list, then passes the trimmed list to `split_atom_words`, which uses `pos[0]` as the origin for byte offsets into the atom's content. Dropping the leading span leaves offset 0 pointing at the wrong line, so every word is reported one line too far down. A Ruby `heredoc_body` always has an empty first span, because the body starts with a newline — so heredocs hit this every time: the whole heredoc is marked changed, highlight runs end at the previous line's length, and the `end` after the heredoc is overdrawn. The same stale span indexes past the display line vector, panicking `--display side-by-side-show-both`. Ordinary string literals escape it because their first line holds the opening quote. tree-sitter-ruby parses these heredocs correctly; the bug is on difftastic's side.

# How

`split_atom_words` now receives the untrimmed spans, so its offset arithmetic starts from the atom's real first line, and trims internally on the path that reports whole lines rather than words. A second commit fixes a related misclassification: difftastic infers "comment" from tree-sitter's `is_extra()`, but Ruby lists `heredoc_body` in `extras` because heredoc bodies occur out of source order, so heredocs were styled as comments and discarded by `--ignore-comments`. That heuristic now yields to the parser's highlight query when the query marks the node as a string. Includes a regression test for an atom with a zero-width leading span.

Written with AI assistance (Claude Code).

### DIFF
--- a/src/parse/syntax.rs
+++ b/src/parse/syntax.rs
@@ -733,6 +733,13 @@ pub(crate) struct MatchedPos {
 ///
 /// If there is negligible text in common with `opposite_content`,
 /// treat the whole `content` as a single novel region.
+///
+/// `pos` and `opposite_pos` must be the *untrimmed* line spans of
+/// their atoms, because offset 0 of `content` is resolved relative to
+/// `pos[0]`. Trimming zero-width spans off the front (as
+/// `filter_empty_ends` does) shifts every word onto the wrong line.
+/// This matters for Ruby heredoc bodies, whose first line span is
+/// always empty because the body starts with a newline.
 fn split_atom_words(
     content: &str,
     pos: &[SingleLineSpan],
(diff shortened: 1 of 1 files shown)


Wilfred:
These are two separate things, can you split them? I agree that the highlighting queries should win over extras for deciding string vs comment. The second case for handling empty regions feels like it might need a fix elsewhere.

Wilfred (changes requested):
See my comment above.

alagos:
I've removed the similar https://github.com/Wilfred/difftastic/commit/843160a8e55c9c1f808cbad9ff99b79ec898902d introduced commit.

PR #1053, opened 2026-09-01
Split class-attribute strings into diffable segments (Elm, HTML, JSX/TSX)

perkee:
>[!WARNING]
> This is not the right way to implement this, and I know that. It's the most straightforward way, though, with only one repo needing changes. Putting it into a PR just so that I can share the diff. This will never leave draft status.

## Summary

Utility-class frameworks like Tailwind pack a lot of meaning into a single string (`class="tw-border-l-500"` -> `class="tw-border-s-500"`), but today difftastic treats the whole string as one atom, so the entire attribute value highlights as changed even when only one segment actually changed.

This adds a tiny vendored `class_list` tree-sitter grammar that splits whitespace-separated tokens, and further splits each token on `-`/`:`/`/`, so class-name changes diff at the segment level instead of the whole string:

```
class="tw-border-l-500"  ->  class="tw-border-s-500"
```
now highlights just `l`/`s`, not the whole class name.

It's wired in via difftastic's existing tree-sitter sub-language injection mechanism (the same mechanism already used for `<script>`/`<style>` in HTML) for:
- **Elm**: `class`/`Html.Attributes.class "..."` calls (matches on the call target's last identifier segment, so both bare and qualified forms work)
- **HTML**: `class="..."` attributes
- **JSX/TSX**: `className="..."` attributes

Only literal string values are matched — dynamically built class lists (`clsx(...)`, template literals, etc.) are out of scope, consistent with how Tailwind's own static analysis requires literal strings anyway.

## Implementation notes

- `parse_subtrees` now honors tree-sitter query predicates (`#eq?`/`#match?`) via `QueryMatch::satisfies_text_predicates`, and looks up the injection target by capture name (`contents`) instead of assuming index 0. This was needed so the new queries can also capture the attribute/function name to filter matches by (e.g. only `class`, not every function call with a single string argument) — the two existing HTML sub-language queries only had a single capture, so this doesn't change external behaviour for them.
- Elm's `string_constant_expr`, HTML's `quoted_attribute_value`, and JS/TS's `string` are no longer forced into `atom_nodes`. Each was safe to un-atomize because their children fully and unambiguously cover the source range (no interpolation-node ambiguity, unlike e.g. JS template strings, which stay atom-forced). Removing HTML's `quoted_attribute_value` also fixes a latent bug: the highlight query targets the inner `attribute_value` node, so attribute values weren't actually getting string-color highlighting before.
- The `class_list` grammar needed a small external scanner (`scanner.c`, same technique already used by the other vendored grammars in this repo) to correctly reject a "dangling" separator (e.g. a trailing `:` in `"hover: foo"`, found in real-world code) as a continuation token — a plain regex can't express "and a segment character follows" since that needs lookahead, which tree-sitter's regex support doesn't have.

## Test plan

- [x] `cargo test` passes
- [x] `cargo fmt --check` clean
- [x] Verified against ~1,200 real HTML/TSX/JSX files and 211 real changed Elm files across several production codebases: zero parse-error regressions and zero crashes compared to an unmodified `master` build
- [x] Manually verified Elm, HTML, and TSX cases highlight only the changed class-name segment, and that unrelated strings (non-`class` function calls, plain `.ts` strings) are unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


perkee:
Retargeting to my fork instead of upstream. This PR is not a good PR for upstream! I'm sorry for the mishegas!

PR #1056, opened 2026-09-03
feat: add nu language

kjubybot:
Adds support for the language of [nushell](https://www.nushell.sh/)


Wilfred:
I'd like to keep new parsers to only dependencies that are on crates.io, is there a released version?

Wilfred:
Feel free to reopen when it's released :)

PR #1058, opened 2026-09-07
Keep metadata-only changes visible with --skip-unchanged

### LINKED ISSUE #1028: --skip-unchanged ignores file permissions changes
I just discovered this while playing around with #1027, somewhat related. The `--skip-unchanged` flag appears to ignore file permission changes, only covering source code changes. This seems like undesired behaviour.

```bash
$ mkdir foo bar foo/baz bar/baz
$ touch foo/foo.txt foo/bar.txt foo/baz/bazinga.txt bar/foo.txt bar/bar.txt
$ chmod +x foo/foo.txt
$ difft --skip-unchanged foo/ bar/
$ difft foo/ bar/
baz/bazinga.txt --- Text
No changes.

bar.txt --- Text
No changes.

foo.txt --- Text
File permissions changed from 100755 to 100644.
No changes.

$ difft --skip-unchanged foo/foo.txt bar/foo.txt
$ difft foo/foo.txt bar/foo.txt 
foo.txt --- Text
File permissions changed from 100755 to 100644.
No changes.

```

logarithmone1128:
Closes #1028.

`--skip-unchanged` drops a file whose permissions changed, printing nothing at all. The same happens for a rename, which #1028 does not mention but is the same cause.

Before, on `master`:

```console
$ GIT_EXTERNAL_DIFF=difft git diff HEAD~1 HEAD
new.txt --- Text
File permissions changed from 100644 to 100755.
No changes.

$ GIT_EXTERNAL_DIFF="difft --skip-unchanged" git diff HEAD~1 HEAD
$
```

and for a rename with identical contents:

```console
$ GIT_EXTERNAL_DIFF=difft git diff -M HEAD~1 HEAD
new.txt --- Text
Renamed from old.txt to new.txt
No changes.

$ GIT_EXTERNAL_DIFF="difft --skip-unchanged" git diff -M HEAD~1 HEAD
$
```

Both descriptions live in `extra_info`, and `print_diff_result` returned early whenever there were no syntactic changes, without looking at it. The binary branch a little further down already prints when only the bytes differ, so this uses the same shape for the text branch.

After, both cases print again, and the flag still does its job:

| | before | after |
|---|---|---|
| permission change | nothing | header + `File permissions changed…` |
| rename, same contents | nothing | header + `Renamed from…` |
| genuinely identical files | nothing | nothing |
| content changed | diff | diff |
| without `--skip-unchanged` | unchanged | unchanged |

Two tests in `tests/cli.rs`. `skip_unchanged_still_reports_a_rename` fails on `master` — I checked by reverting only the `src/main.rs` change and keeping the tests. `skip_unchanged_drops_a_genuinely_unchanged_file` passes either way on purpose: it is there to catch the opposite mistake, of making the flag stop skipping things.

`cargo test` passes, 25 tests.

I left the JSON output alone. `print_directory` filters on `Status::Unchanged` and has the same gap, but JSON is behind `DFT_UNSTABLE` and I did not want to change the shape of an unstable interface without asking. Say the word and I will do that one too.

No changelog entry, per AGENTS.md.


Wilfred:
See AI_POLICY.md

Issue #857, opened 2025-07-05
bug: incorrect/confusing inline diff from rust?


sourcefrog:
I have some changes in Rust source code that are incorrectly (or at least confusingly) represented in difftastic's inline view. It's omitting some unchanged context lines which tends to make them look as if they're being deleted.

<img width="1168" height="455" alt="Image" src="https://github.com/user-attachments/assets/f6f6598f-0328-491d-8656-c792e35a8e4b" />

It looks like the `backup` struct member is being replaced, but it's not. A regular GNU diff makes this clearer:

```diff
--- conserve-before.rs	2025-07-05 10:26:23.983726899 -0700
+++ conserve-after.rs	2025-07-05 10:26:12.390702623 -0700
@@ -105,15 +105,28 @@
     Delete {
         /// Archive to delete from.
         archive: String,
+
         /// Backup to delete, as an id like 'b1'. May be repeated with commas.
-        #[arg(long, short, value_delimiter = ',', required(true))]
+        #[arg(
+            long,
+            short,
+            value_delimiter = ',',
+            required_unless_present = "expiry_days"
+        )]
         backup: Vec<BandId>,
+
+        /// Delete backups that are older than the specified number of days.
+        #[arg(long, required_unless_present = "backup")]
+        expiry_days: Option<u32>,
+
```

Here's the difft output as text:
```
; difft --display=inline conserve-before.rs conserve-after.rs 
conserve-after.rs --- 1/4 --- Rust
105        Delete {
106            /// Archive to delete from.
107            archive: String,
108            /// Backup to delete, as an id like 'b1'. May be repeated with commas.
109            #[arg(long, short, value_delimiter = ',', required(true))]
   110         #[arg(
   114             required_unless_present = "expiry_days"
   118         /// Delete backups that are older than the specified number of days.
   119         #[arg(long, required_unless_present = "backup")]
   120         expiry_days: Option<u32>,
   121 
   122         /// Don't actually delete, just check what could be deleted.
   123         #[arg(long)]
   124         dry_run: bool,

```

The side-by-side view makes it clear the `backup` field isn't actually deleted:

<img width="1525" height="598" alt="Image" src="https://github.com/user-attachments/assets/5da20d14-fff9-463c-b64b-7c407a07e636" />

```
; difft  conserve-before.rs conserve-after.rs
conserve-after.rs --- 1/4 --- Rust
107         archive: String,                                                107         archive: String,
...                                                                         108 
[...]

PR #1018, opened 2026-07-22
Show unchanged lines that fall inside an inline hunk

DivvyCr:
Follow-up to PR https://github.com/Wilfred/difftastic/pull/980 to close Issues https://github.com/Wilfred/difftastic/issues/349 , https://github.com/Wilfred/difftastic/issues/704 , https://github.com/Wilfred/difftastic/issues/978

PR #1059, opened 2026-09-09
Fix inline display

Aurze:
Making inline display better.

As described here: https://github.com/Wilfred/difftastic/issues/857

Fixes:
 - Display both line number
 - contect between changes is properly shown
 - if we have a "sub-line" changes we don't display both lines

Before:
image

After: 
image


Aurze:
Looking into the test (in my defense `cargo test` is working XD)

Aurze:
Noticed an issue with some diff, looking into it

DivvyCr:
I've implemented this as a fully [alternative display mode on my fork](https://github.com/DivvyCr/difftastic/tree/display-inline-interleaved-word), as mentioned in [a comment](https://github.com/Wilfred/difftastic/pull/1018#issuecomment-5293880332) on a preliminary PR to close some simpler inline display issues.

I'm not just looking for @Wilfred to take notice of these seemingly very desired changes...

